PR #9, opened 2015-12-24
Add Middleware Support

Ben-G:
This PR adds support for redux compatible middleware. For now this middleware is provided when initializing `MainStore`.

- [X] Middleware can transform actions, and dispatch new actions with same behavior as redux
- [X] Tests for dispatching within a middleware
- [X] Restructure `MainStore` so that convenience `dispatch` functions play nicely with the dispatch wrapped by the middleware
- [X] ~~Provide API for expecting the combined middleware to return a specific type? Would be useful for promise middleware, etc.~~ For now `dispatch` will return `Any`, allowing middleware to change the return type
- [X] Add Middleware Documentation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,8 @@
 #
 # gitignore contributors: remember to update Global/Xcode.gitignore, Objective-C.gitignore & Swift.gitignore
 
+.DS_Store
+
 ## Build generated
 build/
 DerivedData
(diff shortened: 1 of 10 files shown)


Ben-G:
Until now `dispatch(action: ActionType, callback: DispatchCallback?)` was the designated dispatch method. It needs to correctly use the wrapped dispatch internally.

The wrapped dispatch now needs to take the role of the designated dispatch method, and e.g. use `dispatch_async` to queue action handling at the end of the run loop.

russbishop:
FYI: `swift-lint` fails so carthage can't build off this branch by default.

Ben-G:
@russbishop Interesting & thanks for letting me know! `swift-lint` is only producing a few warnings for me, but no errors. Is it causing build errors for you when you check out this branch?

PR #26, opened 2016-01-03
Remove reduntant constructor from Store protocol

garnett:
Simplifies public interface of the ```Store``` and removes kind of duplication in ```MainStore``` implementation.

The tests are passing, CI build is broken due carthage/github issues.

### DIFF
--- a/SwiftFlow/CoreTypes/MainStore.swift
+++ b/SwiftFlow/CoreTypes/MainStore.swift
@@ -31,19 +31,11 @@ public class MainStore: Store {
     private var subscribers: [AnyStoreSubscriber] = []
     private var isDispatching = false
 
-    public required init(reducer: AnyReducer, appState: StateType) {
+    public required init(reducer: AnyReducer, appState: StateType, middleware: [Middleware] = []) {
         self.reducer = reducer
         self.appState = appState
-        self.dispatchFunction = self._defaultDispatch
-    }
-
-    public required init(reducer: AnyReducer, appState: StateType, middleware: [Middleware]) {
-        self.reducer = reducer
-        self.appState = appState
-        self.dispatchFunction = self._defaultDispatch
-
         // Wrap the dispatch function with all middlewares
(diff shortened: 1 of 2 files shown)


garnett:
@Ben-G if this change makes sense for you, do you think it is worth it to remove 
   ``` func dispatch(action: Action) -> Any``` and friends and leave only extended versions such as ```    func dispatch(action: Action, callback: DispatchCallback?) -> Any``` ?

It will simplify public interface of ```Store``` protocol, I do not see any real consequences for end user of ```MainStore```, because it's implementation will provide default ```nil``` value for ```callback``` parameter. As a result, 'real' public API won't change

Ben-G:
Hey @garnett! Thanks a lot for your input on this. 

> Simplifies public interface of the Store and removes kind of duplication in MainStore implementation

I actually want to keep both initializers here. It's important to indicate that a store can also be created without a middleware. You're absolutely right regarding the redundancy in the initializer though! The implementation was sloppy, I've opened a PR that addresses that #29.

> @Ben-G if this change makes sense for you, do you think it is worth it to remove 
func dispatch(action: Action) -> Any and friends and leave only extended versions such as func dispatch(action: Action, callback: DispatchCallback?) -> Any ?
>
It will simplify public interface of Store protocol, I do not see any real consequences for end user of MainStore, because it's implementation will provide default nil value for callback parameter. As a result, 'real' public API won't change

Right now there's only one `Store` implementation, the `MainStore`. But there likely will be more in future,  I've been playing around with some in local branches. *Therefore I want the public interface defined in `Store` to include all the methods that all stores should provide*. This includes the overloaded versions of dispatch.

The API would look nicer, if it was possible to provide default arguments as part of protocol method definitions, but that isn't possible.

I'll leave this open for discussion, but I think we shouldn't make this change.

garnett:
@Ben-G I see your point here. I have no strong arguments here, you can either close this PR or wait for someone's else input.

PR #28, opened 2016-01-03
Add super calls to sample code

garnett:
I understand that the goal of this sample is to provide an idea how to use store, but nevertheless I've found missing super calls somehow confusing and decided to add them


Ben-G:
Yes, they definitely should be there! Thanks!

PR #32, opened 2016-01-04
Ignore multiple identical subscribers

garnett:
I had to make ```subscribers``` internaly visible in order to test it.

### DIFF
--- a/SwiftFlow/CoreTypes/MainStore.swift
+++ b/SwiftFlow/CoreTypes/MainStore.swift
@@ -38,10 +38,9 @@ public class MainStore: Store {
     public required init(reducer: AnyReducer, appState: StateType, middleware: [Middleware]) {
         self.reducer = reducer
         self.appState = appState
-        self.dispatchFunction = self._defaultDispatch
 
         // Wrap the dispatch function with all middlewares
-        self.dispatchFunction = middleware.reverse().reduce(self.dispatchFunction) {
+        self.dispatchFunction = middleware.reverse().reduce(self._defaultDispatch) {
             dispatchFunction, middleware in
                 return middleware(self.dispatch, { self.appState })(dispatchFunction)
         }
--- a/SwiftFlow/CoreTypes/MainStore.swift
+++ b/SwiftFlow/CoreTypes/MainStore.swift
@@ -28,7 +28,7 @@ public class MainStore: Store {
     public var dispatchFunction: DispatchFunction!
 
     private var reducer: AnyReducer
-    private var subscribers: [AnyStoreSubscriber] = []
+    var subscribers: [AnyStoreSubscriber] = []
     private var isDispatching = false
 
     public required convenience init(reducer: AnyReducer, appState: StateType) {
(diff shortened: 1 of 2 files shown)


Ben-G:
Thanks @garnett! This looks great! As a last step, could you look into the style violations that hound is mentioning? These are coming up, because some of the code violates our [SwiftLint](https://github.com/realm/SwiftLint) ruleset.

The easiest way to do that, is to install SwiftLint locally; that will cause all style violations to show up as warnings in Xcode. Then they should be easy to fix.

Would be great if you could commit a fixup for that, then we're good to go.

Thanks a lot for your contribution :100:

garnett:
@Ben-G done!

Ben-G:
Thanks a lot! I'll need to look into this travis issue before merging, potentially together with @agentk. I can see that a PR on the ReduxKit repo has the same issue.

The underlying issue is that Travis CI does not export environment variables in "untrusted' builds - e.g. forks from an other repo for security reasons:

> Similarly, we do not provide these values to untrusted builds, triggered by pull requests from another repository.

This means that for untrusted builds, the `GITHUB_ACCESS_TOKEN` is not exported; therefore carthage runs without an access token and runs into the API rate-limit issue. This is only an issue for binary downloads - carthage skips binaries when it can't download them and fetches the repos; but Nimble currently doesn't build from source with Carthage on tvOS, so we need to use the binary for that dependency...

Kind of tricky! 

We could:

- Move away from Carthage as main dependency manager (though that seems overblown since it's a temporary issue which should be resolved once the issue with Nimble is fixed)
- Check in the Carthage dependencies
- Other ideas?

Let me sleep over this and maybe I can discuss it with @agentk who is the CI setup expert - I definitely want to get this merged tomorrow though (PST).

Sorry for the delay!

agentk:
Correct. I'm looking into it for https://github.com/ReduxKit/ReduxKit/pull/27 too. I'm looking at using carthage to download the source, then a manual build script for nimble and quick.

Ben-G:
@agentk Fantastic!

agentk:
Btw, the environment variable shouldn't normally be an issue with nimble and quick because just the binaries are downloaded and not done over a pull request. It looks like something has changes though.

Ben-G:
In my understanding carthage only uses the access token for binary downloads, [see] here(https://github.com/Carthage/Carthage/pull/669/files). Because cloning a public repository can be done via git, without the need for an access token.

This means only the PRs fail since they don't get access to the token, right?

agentk:
That makes more sense. I was right, if by right, we mean wrong. :-)

agentk:
I have not found a reliable way of building Nimble-tvOS using Carthage yet. In order to get builds going again in ReduxKit I've changed the travis tvos target from 'test' to 'build' and changed carthage to build from source on ios and mac only. I'll drop the same into a Swift-Flow fork. You can merge it if your happy enough with it.

Ben-G:
Thanks a lot for your contribution @garnett!

And thanks a lot @agentk for fixing this!

Ben-G:
@garnett I would like to follow similar [guidelines as Moya, when it comes to contributions](https://github.com/Moya/contributors). So I've invited you as a collaborator of the organization. 

No pressure to accept!

PR #34, opened 2016-01-05
[Travis] Change Travis tvos checks to build only

agentk:
Change Travis to compile carthage dependencies from source, and only for iOS and Mac.
Change Travis tvOS check from test to build.


Ben-G:
Thanks a lot :+1:

PR #40, opened 2016-01-17
Add `PartialReducer` protocol

mosamer:
This is to allow breaking down complex App State structs into smaller components easier to handle by multiple reducers.

### DIFF
--- a/SwiftFlow/CoreTypes/Reducer.swift
+++ b/SwiftFlow/CoreTypes/Reducer.swift
@@ -25,3 +25,23 @@ extension Reducer {
     }
 
 }
+
+protocol PartialReducer: Reducer {
+    
+    typealias PartialStateType: StateType
+    typealias HandledActionType: Action
+    
+    func get(state: StateType) -> PartialStateType
+    func set(partialState: PartialStateType, originalState: StateType) -> StateType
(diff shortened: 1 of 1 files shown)


Ben-G:
Hi @mosamer, thanks a lot for your contribution!

I understand the idea behind your implementation, but Swift Flow actually already allows you to break down the app state nicely into different slices that are handled by different reducers.

You can see an example of it here: https://github.com/Swift-Flow/GitHubBrowserExample/blob/master/SwiftFlowGitHubBrowser/AppReducer.swift

I like this pattern a lot, since it works with the existing, simple code base.

I believe this should solve the problem you are trying to tackle, without the need for additional code! I'll leave this open for discussion.

Thanks again :+1:

agentk:
Partial reducer functionality is also well handled by reducer composing and AnyReducer type erasing.
Any reducer can call other reducers with fragments of the state to reduce actions to making reducers easy to test in isolation. Fictional example to illustrate reducer composability structure:

```swift
func appReducer(state: State, action: Action) -> State {
    return State(
        auth: authReducer(state.auth),
        records: recordsReducer(state.records))
}
```

Ben-G:
It seem like we are in agreement that this change is not necessary. I'll close the PR. Happy to reopen if discussion is required :smiley:

PR #42, opened 2016-01-18
Type Related Store Improvements

Ben-G:
- Action Creators now Receive Typed Reference to Store, this fixes an issue where it was necessary to cast the `store` argument in an action creator in order to be able to use `dispatch
- Renames `Store` to `StoreType` and `MainStore` to `Store`
- The initial state can be `nil` which makes it possible for reducers to provide the initial state (thanks to @Aleksion and @agentk, sticking with the Redux way is much better here :))


SwiftFlow/CoreTypes/Store.swift
```diff
@@ -1,216 +1,135 @@
...
+        } else {
+            dispatch(SwiftFlowInit())
+        }
+    }
 
-     ```
-     store.dispatch( noteActionCreatore.deleteNote(3) )
-     ```
+    public func subscribe(subscriber: AnyStoreSubscriber) {
+        guard subscribers.indexOf({ $0 === subscriber }) == nil else {
```
dcvz: Is there a reason this is preferred over
```swift
if subscribers.contains({ $0 === subscriber }) { ... }
```
I believe it's a bit more readable and I feel like you'd use indexOf only when you plan to use the resulting index

SwiftFlow/CoreTypes/Store.swift
```diff
@@ -1,216 +1,135 @@
...
+            subscriber._newState(state)
+        }
+    }
 
-     - returns: By default returns the dispatched action, but middlewares can change the
-     return type, e.g. to return promises
-     */
-    func dispatch(actionCreator: ActionCreator, callback: DispatchCallback?) -> Any
+    public func unsubscribe(subscriber: AnyStoreSubscriber) {
+        let index = subscribers.indexOf { return $0 === subscriber }
```
dcvz: maybe a more concise way would be
```swift
if let index = subscribers.indexOf({ return $0 === subscriber }) {
    subscribers.removeAtIndex(index)
}
```

PR #44, opened 2016-01-19
Use GetState alias

sendyhalim:
I think using alias `GetState` would make applying middlewares more readable :relieved:

### DIFF
--- a/Cartfile.resolved
+++ b/Cartfile.resolved
@@ -1,2 +1,2 @@
-github "Quick/Nimble" "v3.0.0"
+github "Quick/Nimble" "v3.1.0"
 github "Quick/Quick" "v0.8.0"
--- a/ReSwift/CoreTypes/Action.swift
+++ b/ReSwift/CoreTypes/Action.swift
@@ -17,7 +17,7 @@ import Foundation
 
  It is recommended that you define your own types that conform to `Action` - if you want to be able
  to serialize your custom action types, you can implement `StandardActionConvertible` which will
- make it possible to generate a `StanardAction` from your typed action - the best of both worlds!
+ make it possible to generate a `StandardAction` from your typed action - the best of both worlds!
 */
 public struct StandardAction: Action {
     /// A String that identifies the type of this `StandardAction`
--- a/ReSwift/CoreTypes/Middleware.swift
+++ b/ReSwift/CoreTypes/Middleware.swift
@@ -9,5 +9,5 @@
 import Foundation
 
 public typealias DispatchFunction =  (Action) -> Any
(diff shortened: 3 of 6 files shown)


Ben-G:
Hey @sendyhalim. Thanks a lot for this PR! 

At first I was: "this change looks great, a lot more readable than the closure solution".

Then I realized that both the original version and the new one create a retain cycle :disappointed:. Since we're passing a strong reference of an instance method to the middleware, the middleware will now retain the store, while the store will also retain the middleware. 

This means we'll need to do something along the lines of:

```
return middleware(self.dispatch, { [weak self] in self?.getState })(dispatchFunction)  
```

Which unfortunately would make the code less readable again.

I think we could use your idea to still make the code more readable:

```
let getState = { [weak self] in self?.state }
return middleware(self.dispatch, getState)(dispatchFunction)
```

This way we don't need the additional instance methods but still achieve your readability improvement :smile: 

What are your thoughts on that solution?

sendyhalim:
@Ben-G oh yea about the strong reference, yep I agree I totally missed that one. Yep the only solution is to create a weak reference, I'll push the commit again. Thanks for the feedback!

Ben-G:
Awesome! I will merge #42 into master in just a moment which has some larger changes. You might want to wait until that is done and then rebase or create a new branch from master. Should make merging easier :smile:

sendyhalim:
@Ben-G anyway isn't the `self` inside the `dispatchFunction` also retain a strong cycle?
Shouldn't it be like this?
```
self.dispatchFunction = middleware.reverse().reduce(self._defaultDispatch) {
    [weak self] dispatchFunction, middleware in
        let getState = { self?.state }
        return middleware(self?.dispatch, getState)(dispatchFunction)
}
```

Ben-G:
@sendyhalim Yes! You are absolutely correct. Good catch :smile: 

What you suggested looks good to me! Also #42 is now merged into master. So you should be able to rebase this branch with the master branch and then make the suggested change!

sendyhalim:
@Ben-G done :D! Let me know if there's any feedback for the PR

Ben-G:
@sendyhalim Thanks a lot! This looks great! :fireworks:.

It's a tiny bit unfortunate that we need to optionally chain in the middleware - but I can't think of a better, clean solution right now.

Let's merge this and revisit that later :smile:.

Our policy is also to invite anyone who has contributed, as a collaborator to the project. Feel free to ignore the invite if you don't want to join :smile:

sendyhalim:
@Ben-G thanks for the invitation, accepted :D

PR #46, opened 2016-01-20
Refactor code to be more concise and readable

dcvz:
- changed signature on `handleAction:StateType:Action` to take `Action` as the first parameter. It doesn't make sense to have an unnamed `StateType` be the first parameter when the method name is `handleAction`
- removes uses of `self` when not necessary as the use of it was on and off throughout the code
- shortened some code
    - replaces the use of `indexOf` for `contains` in places where it doesn't make sense to use
    - shortens some `if let` cases
- replace internal `withSpecificTypes:` with `castToExpectedType:` which name I believe makes more sense and changed the return type to an optional of the expected type. This makes the case where it can't be casted be handled by the caller.


ReSwift/CoreTypes/MainStore.swift
```diff
@@ -38,14 +38,14 @@ public class MainStore<State: StateType>: Store {
         self.state = state
 
         // Wrap the dispatch function with all middlewares
-        self.dispatchFunction = middleware.reverse().reduce(self._defaultDispatch) {
-            dispatchFunction, middleware in
-                return middleware(self.dispatch, { self.state })(dispatchFunction)
+        self.dispatchFunction = middleware.reverse().reduce(_defaultDispatch) {
+            function, middleware in
+                return middleware(self.dispatch, { self.state })(function)
         }
```
Ben-G: `dispatchFunction` is more expressive than `function`. Swift in general favors expressiveness over brevity. I would prefer if this remained the existing way.

ReSwift/CoreTypes/MainStore.swift
```diff
@@ -70,11 +68,10 @@ public class MainStore<State: StateType>: Store {
...
 
         isDispatching = true
-        let newState = self.reducer._handleAction(self.state, action: action)
+        let newState = reducer._handleAction(action, state: state)
         isDispatching = false
 
-        self.state = newState as! State
-
+        state = newState as! State
         return action
```
Ben-G: I've *tried* to be consistent, and keeping an empty line before `return` statements that actually return a value. I think that makes the code easier to parse.

ReSwift/CoreTypes/MainStore.swift
```diff
@@ -91,28 +88,23 @@ public class MainStore<State: StateType>: Store {
...
         return returnValue
     }
 
     public func dispatch(actionCreatorProvider: ActionCreator, callback: DispatchCallback?) -> Any {
-        let action = actionCreatorProvider(state: self.state, store: self)
-        if let action = action {
-            dispatch(action, callback: callback)
-        }
+        let action = actionCreatorProvider(state: state, store: self)
+        _ = action.map { dispatch($0, callback: callback) }
```
Ben-G: Here I strongly prefer the optional chaining version over `map`. It is easier to read and using `map` for side-effects (like here) should generally be avoided.

ReSwift/CoreTypes/MainStore.swift
dcvz: this was from a previous change I'd been experimenting with earlier, i'll change it back.

ReSwift/CoreTypes/MainStore.swift
```diff
@@ -91,28 +88,23 @@ public class MainStore<State: StateType>: Store {
...
 
     public func dispatch(actionCreatorProvider: AsyncActionCreator, callback: DispatchCallback?) {
-        actionCreatorProvider(state: self.state, store: self) { actionProvider in
+        actionCreatorProvider(state: state, store: self) { actionProvider in
             let action = actionProvider(state: self.state, store: self)
-            if let action = action {
-                self.dispatch(action, callback: callback)
-            }
+            _ = action.map { self.dispatch($0, callback: callback) }
         }
```
Ben-G: See note above :sweat_smile:

ReSwift/CoreTypes/CombinedReducer.swift
```diff
@@ -27,10 +27,9 @@ public struct CombinedReducer: AnyReducer {
         self.reducers = reducers
     }
 
-    public func _handleAction(state: StateType, action: Action) -> StateType {
+    public func _handleAction(action: Action, state: StateType) -> StateType {
         return reducers.reduce(state) { currentState, reducer in
-            reducer._handleAction(currentState, action: action)
+            reducer._handleAction(action, state: currentState)
         }
     }
```
Ben-G: :100 in deed much better!

ReSwift/CoreTypes/MainStore.swift
```diff
@@ -38,14 +38,14 @@ public class MainStore<State: StateType>: Store {
...
+                return middleware(self.dispatch, { self.state })(function)
         }
     }
 
     public func subscribe(subscriber: AnyStoreSubscriber) {
-        guard subscribers.indexOf({ $0 === subscriber }) == nil else {
+        if subscribers.contains({ $0 === subscriber }) {
             print("Store subscriber is already added, ignoring.")
             return
         }
```
Ben-G: Nice :+1:

ReSwift/Utils/TypeHelper.swift
```diff
@@ -9,14 +9,12 @@
...
-    action: Action, @noescape function: (state: SpecificStateType, action: Action)
-    -> SpecificStateType) -> StateType {
-
-    guard let specificStateType = state as? SpecificStateType else { return state }
+ Tries to cast a generic `StateType` to the specific type expected by reducers / store subscribers.
 
-    return function(state: specificStateType, action: action) as! StateType
+ **NOTE: Method is only used internally.**
+*/
+func castToExpectedType<SpecificStateType, Action>(state: StateType, action: Action,
```
Ben-G: :+1: Much better naming

Ben-G:
Thanks @dcvz a lot of great stuff in here :100: :+1:!

I've added some comments on changes that I don't entirely agree with - very happy to discuss if you have a different opinion :)

Some more notes:

- **Dependence on other PRs**: Sorry about the unfortunate timing for this PR; it comes between the renaming of the project and PR #42 which renames some of the types that you have refactored in here. We'll have to merge #42 first and then look to rebase your change - after looking into which individual changes we want it might even make sense to just create a new commit from the new master branch
- **Hound Issues**: Would be great to look into the issues hound reported, we've so far managed to keep the 100 character line length in tests, and it's been super helpful when reading the code on small screens. The easiest way to look into these issues is to install SwiftLint locally and ensure that no additional linter warnings were generated (we currently have 1 for a TODO)
- **Test Coverage**: This PR brings the code coverage < 100%, this is due to [a bug in the Swift Compiler](https://bugs.swift.org/browse/SR-420). Some of the `if let` statements need to be unrolled in order to avoid this bug. In my opinion we should keep doing this for now - so it is easy to detect if test coverage dropped. Once it's below 100% it's easier to let untested changes slip by mistake - but I'm open to discuss this one

Sorry for the long prose! Looking forward to getting your contributions in :fireworks:

ReSwift/Utils/TypeHelper.swift
dcvz: do you know why it's giving the invalid alert?

Ben-G:
#42 Merged now :) No outstanding major changes to be merged to master.

PR #47, opened 2016-01-20
Refactor code to be more concise and readable

dcvz:
- changed signature on `handleAction:StateType:Action` to take `Action` as the first parameter. It doesn't make sense to have an unnamed `StateType` be the first parameter when the method name is `handleAction`
- removes uses of `self` when not necessary as the use of it was on and off throughout the code
- shortened some code
    - replaces the use of `indexOf` for `contains` in places where it doesn't make sense to use
    - shortens some `if let` cases


Ben-G:
:star2: :star: Thanks a lot for working through this - great contribution :star: :star2: 

All contributors get invited as collaborators to the org! Accept the invite if you're interested!

PR #53, opened 2016-01-21
Fix Counter Example missing link

sendyhalim:
Shouldn't the counter example points to this https://github.com/ReSwift/CounterExample-Navigation-TimeTravel?


Ben-G:
It should! Thanks a lot for the fix!

sendyhalim:
You're welcome :+1:

PR #55, opened 2016-01-22
Refactor Type helper

sendyhalim:
This PR includes:

- Add trailing newline to TypeHelperTests
- Using 2 `guard` instead of nesting `guard` inside `if` clause
- Use `state` as argument label, use `genericState` for variable name

I don't know if using 2 `guard`s would make it more readable or not, I will revert it if it does not suit the  coding style


Ben-G:
@sendyhalim Very nice change! :+1:  

Would you mind squashing your commits into one? These are small enough changes to go into a single commit, which will make it much easier to browse the history of the project.

sendyhalim:
@Ben-G Sure :tada:

Ben-G:
Thanks a bunch! :ship:

Issue #43, opened 2016-01-19
Substate Selection upon Subscription


Ben-G:
As @Aleksion and I discussed, it would be great to be able to select a sub-state when subscribing to the store. This would allow us to programmatically ensure that the subscriber can only access a subset of the state.

Here's one idea of mine how we could potentially implement it, providing a closure that performs the sub select. We could typecheck that the closure returns the same type as is expected in `newState`.

    override func viewWillAppear(animated: Bool) {
        ...
        store.subscribe(self, { $0.navigationState })
    }

    func newState(state: NavigationState) {
        contacts = state.dataState.contacts
    }

@Aleksion also mentioned another approach using containers outside of View Controllers that would select and pass down a substate to a View Controller explicitly.

raheelahmad:
I would vote for the second model, sort of like dependency injection.

I think there are two other related issues:

1. A subscriber is able to subscribe to any state
2. The state needs to be stored globally. This sounds more problematic to me. There is a tight coupling between state subscribers and this global state variable. Makes testing difficult etc.

Thinking out aloud: For every substate `SubState` we could have a corresponding `SubStateConsumer`

```swift
protocol SubStateConsumer {
    var state: SubState?
}
```

where the ViewController conforms to `SubStateConsumer` and its `state` property is set by the outside "container". The container itself holds the AppState private.

Aleksion:
We were actually discussing an approach not unlike what you're proposing here.
In ReduxJs you have the concept of container components - which are the only components that know about the store and the whole state.

These would then pass down relevant functions and state to it's sub components.

This approach would make it easier to create isolated components - that could potentially be reused across projects. 

I kinda like the opportunities - we just have to figure out how to beat implement it and the potential downsides

Ben-G:
This blog post contains some interesting insight into how Om (current and feature version) do and don't handle state subselects. Should provide some good insights:
http://blog.circleci.com/why-we-use-om-and-why-were-excited-for-om-next/

Ben-G:
@Aleksion @raheelahmad My proposed solution would make substate selection look like this:

```
func selectSubstate(state: TestComplexAppState) -> (Int?, String?) {
        return (
            state.testValue,
            state.otherState?.name
        )
}
```

It allows you to select an arbitrary combinations of paths from your state. I'm actually pretty happy with this. Would be interested in your thoughts!

Ben-G:
This is solved by #61.

PR #60, opened 2016-01-24
Allow Subscribers to Subselect a State

Ben-G:
We have been discussing this in #43 and I think I'm pretty excited about this solution. This PR adds a method called `selectSubstate` to `StoreSubscriber`. 

The method comes with a default implementation, so it doesn't have to be implemented. You can implement it to subselect the state you are interested in. This solves the problem of always having acess to the global app state within a subscriber.

With a glance at `selectSubstate` you will be able to see which part of the state is flowing into a subscriber. I think it's about as declarative as it can get. Here's an example from the tests of what a selecting subscriber would look like:

```
class TestSelectiveSubscriber: StoreSubscriber {
    var receivedValue: (Int?, String?)

    func selectSubstate(state: TestComplexAppState) -> (Int?, String?) {
        return (
            state.testValue,
            state.otherState?.name
        )
    }

    func newState(state: (Int?, String?)) {
        receivedValue = state
    }

}
```

The input is the app state, the return type can be `Any` type, as long as it matches the argument of `newState`.

Waiting for some feedback. If we like this, I'll add some comments and then we can merge :ship:.

### DIFF
--- a/ReSwift.xcodeproj/project.pbxproj
+++ b/ReSwift.xcodeproj/project.pbxproj
@@ -7,6 +7,9 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		250F94D31C548B90001FAAC8 /* StoreSubscriberSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = 250F94D21C548B90001FAAC8 /* StoreSubscriberSpec.swift */; };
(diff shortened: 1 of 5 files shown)


Ben-G:
Ah.. I missed an edge case in testing. I must be getting tired; I could bet that Xcode showed 100% :smile:

Ben-G:
:fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire:
In favor of #61 
:fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire:

PR #61, opened 2016-01-24
Substate Selection via Closure upon Subscription

Ben-G:
:fire: :fire:  #60 :fire: :fire:

Here comes a much more elegant solution. Thanks for @agentk for pushing me in this direction.
You now can optionally provide a closure upon subscription to build a subtree of the state you're interested in:

                store.subscribe(subscriber) {
                    (
                        $0.testValue,
                        $0.otherState?.name
                    )
                }

As an additional bonus, this PR introduces generic constraints on `Subscriber` so that the compiler will now complain if the subscriber is not listening to the `State` type of the store (when no closure is provided) or when the return type of the selector does not match the type the `Subscriber` is interested in.

### DIFF
--- a/ReSwift.xcodeproj/project.pbxproj
+++ b/ReSwift.xcodeproj/project.pbxproj
@@ -14,6 +14,9 @@
 		259737EA1C2C611600869B8F /* StoreMiddlewareTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 259737E91C2C611600869B8F /* StoreMiddlewareTests.swift */; };
 		259737FB1C2C618A00869B8F /* TestFakes.swift in Sources */ = {isa = PBXBuildFile; fileRef = 259737FA1C2C618A00869B8F /* TestFakes.swift */; };
 		259737FD1C2C661100869B8F /* Middleware.swift in Sources */ = {isa = PBXBuildFile; fileRef = 259737FC1C2C661100869B8F /* Middleware.swift */; };
+		25AB3B811C54A84500A41513 /* StoreSubscriberSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = 25AB3B7F1C54A79F00A41513 /* StoreSubscriberSpec.swift */; };
+		25AB3B821C54A84600A41513 /* StoreSubscriberSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = 25AB3B7F1C54A79F00A41513 /* StoreSubscriberSpec.swift */; };
+		25AB3B831C54A84700A41513 /* StoreSubscriberSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = 25AB3B7F1C54A79F00A41513 /* StoreSubscriberSpec.swift */; };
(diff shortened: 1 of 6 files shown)


Aleksion:
Well that's much cleaner!
And the way I see it, it enables passing down substates to viewControllers. So that the viewControllers will only ever receive the state that's relevant to them. 

This is definitely a huge step in the right direction. I look forward to playing around with it.

Ben-G:
Received positive feedback from @Aleksion and @agentk - :ship:

raheelahmad:
I like it. It wasn't obvious to me at first, but you could return any type as the subtree of selection (and not just a tuple). 

For e.g., returning a `struct` makes intention more explicit in the Subscribers: https://github.com/raheelahmad/ReSwift/blob/closure-substate-selection/ReSwiftTests/StoreSubscriberSpec.swift#L60

Ben-G:
@raheelahmad Yes, exactly. Nice example :)

DivineDominion:
As in my comment above and #63, I think here's a Subscription waiting to come out. I was playing with a value type instead of the tuple and found that "substate selection" is basically a `map` rather than a `filter` operation.

For example:

```swift
extension String: StateType {}
class StringSubscriber: StoreSubscriber {
    typealias StoreSubscriberStateType = String
    func newState(state: StoreSubscriberStateType) { }
}

_ = Subscription(subscriber: StringSubscriber()) // works
_ = Subscription(subscriber: StringSubscriber()) { state in
    return state.lowercaseString
}
```

I'm not sure what to do about `AnyStoreSubscriber`, yet, as it doesn't care about a particular `StateType` at all. I think this is weird because why should it be possible to pass in `Any` when the store deals with stuff that extends `StoreType`?

n0quarter:
Hey guys, I'm having error while using the syntax described above with Swift 3:
`mainStore.subscribe(self) {
      $0.participants
    }`

The error is `Cannot convert value of type '(_) -> _' to expected argument type '((_) -> _)?'`

What am I doing wrong ?

danielmartinprieto:
Hey! Declare the function newState with the proper type and that will help Swift infer the type of the closure. Hope it helps.

n0quarter:
Daniel, tnx a lot, this helped.

To whom it may concern, the working code is:
```Swift
mainStore.subscribe(self) {
      (
        $0.participants,
        $0.invitePressed
       )
    }

typealias ParticipantsViewSubState = (participants : [Participant]?, inviteTapped: Bool)

func newState(state: ParticipantsViewSubState) {
  //handle state change
}
```

Also works with Structs or, I assume, any other type.

But I also have more general question related to the topic.
Let's say I have a view, which is interested in a subState of 15 different properties.
But I only want to do `tableView.reloadDate()` if one certain property of subState has changed.

Is this possible anyhow ?

danielmartinprieto:
I think you have several options. You can save the old substate in your view controller and diff it with the new one, updating only the subviews needed, or even better, you can subscribe each view to the store with the part of the substate that view is interested in (you'll need to subclass the table view in this case).

n0quarter:
I was looking for a more elegant solution, ideally provided by the ReSwift :)
Thank you anyway.

nikans:
This should be in the docs

PR #63, opened 2016-01-25
Conceptualize Subscriptions

DivineDominion:
The `Store.subscribers` array was extended as discussed in #60 to include a substate selector. The resulting tuple is not a subscriber itself anymore but rather a subscription. The `typealias` opens up ways to extend this even further with a `struct` replacement later.

### DIFF
--- a/ReSwift/CoreTypes/Store.swift
+++ b/ReSwift/CoreTypes/Store.swift
@@ -21,7 +21,7 @@ public class Store<State: StateType>: StoreType {
 
     /*private (set)*/ public var state: State! {
         didSet {
-            subscribers.forEach {
+            subscriptions.forEach {
                 // if a selector is available, subselect the relevant state
                 // otherwise pass the entire state to the subscriber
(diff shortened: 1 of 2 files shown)


Aleksion:
It's a fairly good point.
I suppose a subscription should be an entity tying together the store and the subscribers subscribing to it.  And since subscribers were changed from being the subscribers themselves, to what you now formulated as the subsription typealias, by that definition you are right.

Your code obviously doesn't break anything, and you've added a protocol that might be neat to have at a later time. So I'm inclined to approve.

@Ben-G - since you formulated the original store and subscribers (though we did the same in ReduxKit), I think you should make the decision.

Ben-G:
This looks good to me, as it expresses the new semantics of subscribing well! :ship: 

Thanks a lot for the contribution :boom: :star2: 

We're inviting all contributors to become members of the organization. Feel free to ignore the invitation if you're not interested!

Aleksion:
Great Work! :+1:

PR #67, opened 2016-01-30
[Refactor] Split out readme into Jazzy guides

agentk:
The contents of the readme are split out into seperate files to make the documentation easier to follow. This introduces duplicate content :-(. Next step is reduce the duplicate content in readme.md and provide links to the jazzy docs.


Ben-G:
Hey Karl, this looks great, thanks a lot :smiley: 

Do you think you could update the image link in the main README.md? So that we have a working Readme until it is updated?

We should also discuss what exactly should remain in the new Readme. I think I would prefer some duplication, so that the Readme on GitHub is fairly expressive without the need to visit a separate link.

Happy to chat more about that!

agentk:
✅ Updated and squashed. Also fixed a couple of links to the getting started guide.

P.s. Only reason I've changed the Assets to img is to keep the generated jazzy docs a bit cleaner, and keep images in markdown working on GitHub too.

With the duplication in README.md, I agree it should still be comprehensive. I'm thinking at the moment of having the generated docs as the README.md split by h1's + anything split out in the Readme folder.

It would remove all duplication, and keep everything working in GitHub too. It will probably be my next priority over in-code docs.

Ben-G:
Thanks a lot :)

Issue #62, opened 2016-01-25
Store subscription promotes retain cycles in view controllers


russbishop:
The natural way to approach store subscription is to subscribe in `viewDidLoad`, then unsubscribe in `deinit`. Because the `Store` holds a strong reference to the VC this promotes a retain cycle. I feel like this is a bear trap that people will repeatedly step in so maybe it is worth discussing potential solutions. 

1. Tell people not to hold a reference to the store, just access it as a global. This seems like a bad idea and bad design.
2. Tell people to hold a `weak` reference to the store. This relies on everyone doing the "right thing" which we already know they don't do.
3. Have the store hold subscribers as a `weak` reference (I have a `WeakBox` that could be used for this). This means putting unsubscribe in the `deinit` should work just fine to keep the subscriber array clean. When dispatching if we encounter a `nil` subscriber we could remove it or assert depending on whether we consider this a programmer error or not.
4. Tell people to subscribe in `viewWillAppear` and unsubscribe in `viewWillDisappear`. 

It might be obvious but I lean heavily toward 3 because it eliminates a headache for users of the library and allows them to rely on the existing UIKit/AppKit mechanisms for managing View Controller lifetimes. My temporary workaround is to unsubscribe from `viewWillDisappear` but that is problematic for VCs that will in fact re-appear and 4 means you can't react to state changes if you aren't on-screen.

I'm happy to submit a PR for this one if there is consensus on fixing it.

Ben-G:
Hm.. I might be missing something. But shouldn't unsubscribing in `deinit` break the strong reference from the `Store` to the VC?

Ben-G:
Ah, got it :smile: `deinit` won't ever be called because the store holds a strong reference to the VC.
I personally haven't run into this since I use option 4. by default.

But this definitely is an issue that should be fixed. Users of the library would not naturally expect the store to keep a strong reference to the subscribers.

A PR would be welcome :smiley:

russbishop:
@Ben-G Because of the strong reference in the store subscription and the VC's `var store: StoreType` property, when the VC is dismissed (say by popping from the nav stack) there is still a live reference so `deinit` is never called.

DivineDominion:
In another app I circumvented issues like these by adhering to the convention of not subscribing/unsubscribing `self` ever, only child components. Same for notifications.

PR #70, opened 2016-02-04
Store now keeps a weak reference to subscribers

vfn:
Transformed the typealias `Subscription` into a `Struct` and added a weak reference to the `subscriber`.

Fixes https://github.com/ReSwift/ReSwift/issues/62

### DIFF
--- a/ReSwift/CoreTypes/Store.swift
+++ b/ReSwift/CoreTypes/Store.swift
@@ -23,6 +23,7 @@ public class Store<State: StateType>: StoreType {
 
     /*private (set)*/ public var state: State! {
         didSet {
+            subscriptions = subscriptions.filter { $0.subscriber != nil }
             subscriptions.forEach {
                 // if a selector is available, subselect the relevant state
                 // otherwise pass the entire state to the subscriber
--- a/ReSwift/CoreTypes/Store.swift
+++ b/ReSwift/CoreTypes/Store.swift
@@ -72,24 +72,18 @@ public class Store<State: StateType>: StoreType {
 
     public func subscribe<S: StoreSubscriber
         where S.StoreSubscriberStateType == State>(subscriber: S) {
-            if !_isNewSubscriber(subscriber) { return }
-
-            subscriptions.append(Subscription(subscriber: subscriber, selector: nil))
-
-            if let state = self.state {
-                subscriber._newState(state)
-            }
+            subscribe(subscriber, selector: nil)
     }
 
     public func subscribe<SelectedState, S: StoreSubscriber
         where S.StoreSubscriberStateType == SelectedState>
-        (subscriber: S, selector: (State -> SelectedState)) {
+        (subscriber: S, selector: (State -> SelectedState)?) {
             if !_isNewSubscriber(subscriber) { return }
 
             subscriptions.append(Subscription(subscriber: subscriber, selector: selector))
 
             if let state = self.state {
-                subscriber._newState(selector(state))
+                subscriber._newState(selector?(state) ?? state)
             }
     }
 
--- a/ReSwift/CoreTypes/Store.swift
+++ b/ReSwift/CoreTypes/Store.swift
@@ -17,6 +17,8 @@ import Foundation
  */
 public class Store<State: StateType>: StoreType {
 
+    typealias SubscriptionType = Subscription<State>
+
     // TODO: Setter should not be public; need way for store enhancers to modify appState anyway
 
     /*private (set)*/ public var state: State! {
@@ -24,7 +26,7 @@ public class Store<State: StateType>: StoreType {
             subscriptions.forEach {
                 // if a selector is available, subselect the relevant state
                 // otherwise pass the entire state to the subscriber
-                $0.subscriber._newState($0.selector?(state) ?? state)
+                $0.subscriber?._newState($0.selector?(state) ?? state)
             }
         }
     }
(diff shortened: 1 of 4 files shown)


agentk:
Hi @vfn,
Changes look good. Exactly how I was planning on ensuring weak subscriber refs.
Before we merge the commit could you add a test for ensuring dereferenced subscribers are removed from the subscriptions array. This will also keep coverage complete.
Once complete just squash the commits to keep the history concise.
Cheers, Karl. P.s. nice to see someone so close (by Aussie distances at least :-)

ReSwift/CoreTypes/Subscription.swift
```diff
@@ -0,0 +1,14 @@
...
+//  Created by Virgilio Favero Neto on 4/02/2016.
+//  Copyright © 2016 Benjamin Encz. All rights reserved.
+//
+
+import Foundation
+
+struct Subscription<State: StateType> {
+    private(set) weak var subscriber: AnyStoreSubscriber? = nil
+    private(set) var selector: (State -> Any)?
+}
```
Ben-G: Can we just make these `let` instead of `private(set)`?

ReSwift/CoreTypes/Store.swift
```diff
@@ -17,14 +17,26 @@ import Foundation
...
     // TODO: Setter should not be public; need way for store enhancers to modify appState anyway
 
     /*private (set)*/ public var state: State! {
         didSet {
-            subscriptions.forEach {
+            var indexesToRemove: [Int] = []
+            for (index, subscription) in subscriptions.enumerate() {
+                guard let subscriber = subscription.subscriber else {
+                    indexesToRemove.append(index)
+                    return
```
Ben-G: The escape route here should be a `continue` and not a `return`. With the current code subscribers after a no-longer-existent subscriber would miss a state update.

I think it would be best to separate these two pieces of logic:
1) Filter out subscriptions that have lost their reference to a subscriber
2) Notify all remaining subscribers

Could look something like this:

```
subscriptions = subscriptions.filter { $0.subscriber != nil }
subscriptions.forEach {
     $0.subscriber!._newState(subscription.selector?(state) ?? state)
}
```
Since the use of the implicit unwrapped optional here is very close to the line that filters out nil-values, I think it is fine to use and shouldn't through off folks reading the code.

Ben-G:
Hey @vfn, great contribution. Thanks a lot!

See my comments for some issues and some suggested changes. When this is ready to merge we should have test cases for:

1. A subscriber that is no longer referenced strongly should become deallocated, even when subscribed to the store
2. Other subscriptions still work expected, even when a subscription is cleaned up during the next notification cycle

It also seems like this here would conflict with #71. The change in #71 is great, but it probably makes sense to combine the changes into this PR, what do you think?

Lastly, if you don't have time/interest to make the necessary changes, you can also re-open a PR against `dev` and we can take the rest of the implementation over from there. Whatever you prefer :+1: 

Thanks a lot for the contribution :fireworks:

vfn:
Hi @Ben-G and @agentk, thank you for reviewing the PR so quickly. I'm working on the fixes and I'll update the PR in the next few hours.

Do you want me to push the fixes as fixup commits and once it all looks good I/we squash them, before merging, or do you prefer to have the PR already updated with the changes squashed?

Ben-G:
I think it makes sense to review first and squash afterwards! That way we
can save you some time :)

Thanks a lot!

--
Benjamin Encz

Find me on twitter: @benjaminencz <https://twitter.com/benjaminencz>

ReSwift/CoreTypes/Store.swift
vfn: @Ben-G, I like how the code is concise with this approach, although I'm not sure if the best approach here is to run the filter on every attempt to notify the subscribers.

Ben-G: If you like we can look into some performance test, but the operation
should be extremely cheap for all use cases I can think of. I'm also open
to other suggestions :) Ideally though I would like to treat these to tasks
separately to make the code easier to understand and to eliminate the risk
of subtle changes that could cause bugs.

--
Benjamin Encz

Find me on twitter: @benjaminencz <https://twitter.com/benjaminencz>

vfn: I'm happy with your suggestion.

I'm leaning towards keeping the optional chaining when accessing the subscriber instead of implicit unwrapping it. There is no cost for the call, and it saves potential crashes caused by subscribers being deinitalized while in the `forEach`. Does that make sense?

Ben-G: Ah yes, that makes more sense. There's indeed no need to implicitly unwrap
here! Optional chaining it is!

--
Benjamin Encz

Find me on twitter: @benjaminencz <https://twitter.com/benjaminencz>

vfn:
@Ben-G @garnett I have updated the PR with the suggested changes, added the tests and also added the commit from PR https://github.com/ReSwift/ReSwift/pull/71 into this PR.

Unfortunatelly I screwed up and created this PR against `master`, instead of `dev`. Once we finish reviewing the stuff here, I can create another PR agains `dev` and we can reference the discussion on this PR here that can be closed afterwards.

Ben-G:
@vfn This looks great now! Thanks. You can go ahead and rebase now :smiley: 

In this state it's fine to merge directly into `master`. We can always add a test case or two later, but IMO coverage is sufficient for now!

vfn:
@Ben-G I've rebase and squashed the fixups. Let me know if you need anything to be done  before merging it

Ben-G:
@vfn This looks great! Thanks a lot for the contribution :star2: :star2: 

As a contributor I'll invite you to the organization. Feel free to ignore the invite if you're not interested.

Cheers!

PR #71, opened 2016-02-04
Removed ‘duplicated’ implementation by allowing nil selector

vfn:


Ben-G:
This looks great! But might conflict with #70, see details on that issue. Would like to hear your opinion!

vfn:
code moved into https://github.com/ReSwift/ReSwift/pull/70

Issue #66, opened 2016-01-29
Store entities do not deinitialize (dealloc) when no longer owned (Memory Leak)


JesseARWhite:
How to reproduce:
1.  Add a deinit callback to the Store class to know that a store has been deallocated
2.  New up a store in memory that is not owned (meaning it should be deallocated on the next run loop)
3.  deinit will not get called at runtime, meaning the store has a memory leak
4.  Comment out the following lines:
        self.dispatchFunction = middleware.reverse().reduce(_defaultDispatch) {
            [weak self] dispatchFunction, middleware in
                let getState = { self?.state }
                return middleware(self?.dispatch, getState)(dispatchFunction)
        }
5.  Re-test, store now deallocates properly

It appears that the dispatch function is ultimately holding a reference to the store object itself, making it impossible for the store to go away and free up the memory it was using if it was using.  Consequently it also holds onto any reducers and subscribers.  Allowing the store to deallocate will allow all referenced objects to deallocate as well.

Ben-G:
Nice catch! I planned on looking into this. I assume the '_defaultDispatch' reference is the cause of the issue. 

Will look into this next week (after vacation), but a fix as PR is appreciated as well :)

Ben-G:
Closed by #72.

PR #72, opened 2016-02-04
Fix Retain Cycle Caused by Store's Dispatch Method

Ben-G:
Fixes retain cycle within Store: #66. Storing a reference to an instance method in an instance member leads to a retain cycle in Swift, so we need to use an explicit closure.

Within the closure it's important not to treat `self` as optional, otherwise the return type of the `dispatch` method gets turned into an optional as well.

Not the nicest code, but as of now I don't know of any alternatives. We can always revisit in future.


Ben-G:
Merging this since @agentk gave :+1: after review!

agentk:
👌👍😉

PR #77, opened 2016-02-12
Misc

vfn:
Two minor updates, one in the README file to update the `handleAction` function, and the second although quite small is a breaking change. It's the rename of `SwiftFlowInit` to `ReSwiftInit`

The diff is big because `dev` is a few commits behind `master`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,9 @@ Your reducer needs to respond to these different action types, that can be done 
 ```swift
 struct CounterReducer: Reducer {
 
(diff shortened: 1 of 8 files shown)


Ben-G:
@vfn Sorry, I did not actively close this PR. I removed the `dev` branch, since we're actually not using it, which automatically closed the PR.

Sorry for the additional effort, but could you reopen against `master`?

Thanks a lot!

PR #79, opened 2016-02-13
Doc update, and initial action renamed

vfn:
Two minor updates, one in the README file to update the handleAction function, and the second although quite small is a breaking change. It's the rename of SwiftFlowInit to ReSwiftInit


Ben-G:
Thank you! :tada:

PR #87, opened 2016-02-25
add codebeat badge

lsolniczek:
Congratulation!!! 3.94 is one of the best score I've seen so far :)


Ben-G:
Hey @lsolniczek,

thanks for the PR. I personally don't think we need this badge in our Readme. Do @agentk or @Aleksion have a different opinion?

Aleksion:
Well, I don't have a particularly strong opinion about it. If it adds increased credibility and our score is great - it might makes sense. 
But I'm also not a fan of sticking too many badges on the page just for the sake of promotion

lsolniczek:
If I may :) I think @Aleksion is right about credibility. Your code is really great according to our algorithm and it is something that you can say proudly :)

agentk:
If anything, I would prefer to remove badges, rather than add them. The way GitHub proxies image URLs can make the project page quite slow to load.

I think codebeat is a very interesting @lsolniczek. I don't see a lot of value in adding another badge to the project though. Because ReSwift is a framework aimed at other developers, it's credibility lies in simple use and readable, understandable code.

The existing badges are not there for credibility either. Test and coverage badges are for the contribution requirements, and the others indicate the released version and platforms supported.

@lsolniczek, has Code Quest thought about supporting GitHub PR check integration? I get the feeling that would be a better fit for a lot of projects.

Aleksion:
I think that settles it then :)

Sent from Outlook




On Fri, Feb 26, 2016 at 12:20 PM -0800, "Lukasz Solniczek" <notifications@github.com> wrote:





If I may :) I think @Aleksion is right about credibility. Your code is really great according to our algorithm and it is something that you can say proudly :)

---

Issue #65, opened 2016-01-29
The new Reducer protocol forcing a reducer to hydrate the state


soapsign:
With the new `Reducer` protocol beings 
```
public protocol Reducer : AnyReducer {
    typealias ReducerStateType
    public func handleAction(action: Action, state: Self.ReducerStateType?) -> Self.ReducerStateType
}
```
It force every reducer observing a particular `ReducerStateType` to hydrate the state. 
If I have a reducer that do not want to hydrate the state or only want to handle the `Action` if the targeted `State` exits, I can't achieve it due to this protocol.
I think we should have a better way to handle this.

Ben-G:
@soapsign, you are bringing up a good point! The current protocol fairly closely resembles the Redux implementation. It has the advantage that you can initialize a Store without the need of initializing the state. In many cases you won't have all information to provide an initial state at that point in time. 

The idea of the current protocol is, that a reducer owns a slice of your state and thus should know best how to initialize it. 

I however agree that this assumption might not always be correct. Could you maybe share your example? That would make the discussion easier. 

I could imagine that it makes sense to offer to initialization paths:

1. Provide initial state and provide reducer that assumes initialized state
2. Don't provide initial state and provide reducer that can hydrate 

Using a separate protocol we could accomplish this. Definitely worth discussing. 

Thanks for bringing this up!

soapsign:
Consider my `AppState` has 2 subState which is `StateA` and `StateB`. Then 'AppState' needs to conform to `StateType`, `HasStateA` and `HasStateB`. For `HasStateB`, the definition is as follow
```
protocol HasStateB {
    var stateB: StateB? { get set }
}
```
It is defined as `StateB?` as this subState should only available for a limited of time for a small part of the application and should be cleaned after use.

Then I have a `StateBReducer`, which should implement `func handleAction(action: Action, state: HasStateB?) -> HasStateB`.
There are a few problem with this requirement. 
First, what should I return If I want to hydrate this subState.
Second, the `Store` call my `Reducer` no matter what `Action` is dispatching. This makes the `Reducer` to hydrate the state as soon as the first `Action` is dispatched.

Maybe I am doing this optional subState in a wrong way, but this is the problem I face when I want to achieve this optional subState.

soapsign:
```swift
struct AppState: StateType, HasStateA, HasStateB {
    var stateA: State
    var stateB: State?
}

struct State {
    var value: String
}

protocol HasStateA {
[...]

PR #90, opened 2016-03-04
Issue #65

gdunkle:
This is a possible solution for issue #65.

By not requiring the reducer to hydrate the state, reducers can more easily use protocols to access slices of the app state that they are interested in.

### DIFF
--- a/ReSwift/CoreTypes/CombinedReducer.swift
+++ b/ReSwift/CoreTypes/CombinedReducer.swift
@@ -29,8 +29,8 @@ public struct CombinedReducer: AnyReducer {
         self.reducers = reducers
     }
 
-    public func _handleAction(action: Action, state: StateType?) -> StateType {
-        return reducers.reduce(state) { (currentState, reducer) -> StateType in
+    public func _handleAction(action: Action, state: StateType?) -> StateType? {
+        return reducers.reduce(state) { (currentState, reducer) -> StateType? in
             reducer._handleAction(action, state: currentState)
         }!
     }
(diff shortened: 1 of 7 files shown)


Ben-G:
@gdunkle Thanks a lot for looking into this and sorry for the late response! 

The problem with your approach is that the `State` remains optional after initialization of the store. This has many issues downstream as every reference to the state needs to check whether we have an initialized state or not. ReSwift somewhat relies on a guaranteed initialized state.

We haven't discussed this on the thread for #65 yet, but in the new recommended way for building reducers, only your top level reducer needs to implement the `Reducer` protocol.

Here's an example: https://github.com/ReSwift/GitHubBrowserExample/blob/master/SwiftFlowGitHubBrowser/AppReducer.swift

This allows you to implement the sub-reducers as plain functions, which also allows you to define an arbitrary interface. This allows you to implement the following:

```
struct AppReducer: Reducer {

    func handleAction(action: Action, state: State?) -> State {
        return State(
            repositories: repositoriesReducer(state?.repositories, action: action),
            // Allows to use nil-coalescing operator to initialize substate
            bookmarks: bookmarksReducer(state?.bookmarks, action: action) ?? []
        )
    }
}
```

This way the reducer function can take in a non-optional state. I hope this makes sense!

Since we have this, I don't think we need to change the interface of the `Reducer` protocol.

I will expand on this in the discussion for #65 to make this clearer.

Thanks a lot for your contribution!

gdunkle:
@Ben-G I really like the solution you presented, makes perfect sense.

PR #107, opened 2016-03-30
[swift-2.2] Update for selector and incrementor changes in Swift 2.2

agentk:
Target Swift 2.2 in Travis
Update Selector syntax with Swift >= 2.2 compatibility
Update incriminator shorthand syntax
Update Action field description to pass tests


Ben-G:
Thanks for tackling this :+1:

PR #111, opened 2016-04-20
Fix an example code of asynchronous operation in the documentation.

mitsuse:
The example code in the doc seems to be wrong. `SetRepositories(repositories: .Loading)` should be dispatched as soon as the network request starts, but the action in the original example is dispatched just after the request is finished.

Does this modification reflect an intention of the documentation correctly?


Ben-G:
@mitsuse Nice catch! 🎉 

As a contributor you'll be invited to the ReSwift organization on GitHub. Feel free to ignore the invite if you're not interested.

mitsuse:
Thanks!

PR #116, opened 2016-05-12
Fix syntax issue

marcinwyszynski:
Hey, CTO of codebeat here. I was reviewing some of the OS files which failed parsing and I found this one. Looks though like the failure was legit and you have a proper syntax error. If you tried to build this code in Swift < 2.2 it would most likely bomb.

Please see the relevant part of the grammar [here](https://developer.apple.com/library/ios/documentation/Swift/Conceptual/Swift_Programming_Language/Declarations.html#//apple_ref/doc/uid/TP40014097-CH34-ID361).

### DIFF
--- a/ReSwift/CoreTypes/StoreType.swift
+++ b/ReSwift/CoreTypes/StoreType.swift
@@ -19,7 +19,7 @@ public protocol StoreType {
     #if swift(>=2.2)
     associatedtype State: StateType
     #else
-    typealias State: StateType
+    typealias State = StateType
     #endif
 
     /// Initializes the store with a reducer and an intial state.


agentk:
Hi Marcin, thank you for diving into the internals.

The syntax is correct and compiles correctly < swift 2.2.

The typealias is declaring that State is "of" a StateType. Where as changing it to an equal sign declares that State "is" a state type, obviating the need for the actual typealias. 

I'll mark this as closed, feel free to reopen if I've read the request incorrectly or not explained it properly. 

- Karl

marcinwyszynski:
I'll let Apple speak.

screen shot 2016-05-12 at 16 07 45

marcinwyszynski:
`associatedtype` works the way you're describing, and according to Apple (see above) `typealias` does not.

the_swift_programming_language__swift_2_2___declarations

agentk:
Oh yeah, the Apple docs only seem to reference the latest swift 2.2 syntax. Below swift 2.2 the syntax is correct with typealias. It was introduced with https://github.com/apple/swift-evolution/blob/master/proposals/0011-replace-typealias-associated.md.

PR #121, opened 2016-05-29
Add part for store initialisation

thomaspaulmann:
The part of store initialisation was missing, which could cause confusion for some developers  who wanna do a quick start - For example me 💩. 

Any recommendations are welcome ✌️

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,21 @@ struct CounterReducer: Reducer {
 ```
 In order to have a predictable app state, it is important that the reducer is always free of side effects, it receives the current app state and an action and returns the new app state.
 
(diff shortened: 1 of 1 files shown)


Ben-G:
@thomaspaulmann Thanks a lot for the contribution 🎉 This looks good to me. At some point we'll need to revisit the docs and restructure some of it; but this is definitely a missing piece in the main readme.

P.S.: We're inviting all contributors to the organization. In case you're not interested, feel free to ignore the invite!

thomaspaulmann:
@Ben-G Yes, the docs have some missing parts and also some old SwiftFlow stuff, but we can do that later. I'm currently working on another example app for ReSwift. At some point, I can transfer the repo to ReSwift.

PR #125, opened 2016-06-27
ReSwift 2.0.0 Release

Ben-G:
Bumping version number to 2.0.0 for next release. Also adding relevant changelog entries.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,19 @@
 #Upcoming Release
 
+#2.0.0
+
+*Released: 06/27/2016*
+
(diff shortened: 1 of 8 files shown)


Ben-G:
*Note on the code change as part of this Release*:

I decided to make on change to StandardAction which handles the serialization of nil payload values. A nil payload is now simply represented by a string constant defined by ReSwift. Technically we should be using NSNull but since ReSwift has no other Foundation dependencies I think we should avoid adding it.

In the end the serialization/deserialization of nil payloads is entirely owned by the library itself so it doesn't really matter how we represent them.

PR #126, opened 2016-07-02
[Swift3] Migrate project to support Swift 3

agentk:
It's time to open the discussion on how to handle Swift 3 migration.

We have a few options.

- Move to Swift 3 only support
- Support Swift 2...3 maintaining Swift 2 naming conventions
- Support Swift 2...3 using Swift 2 naming conventions for users of Swift 2, and Swift 3 naming for Swift 3 users
- Support Swift 2...3 using only Swift 3 naming conventions

It's actually quirt easy to still support Swift 2+ in the one project, and Swift 3/Xcode 8 will not be usable for submitting apps for a few months still, so I'd take the first option of the table.

Then that leaves us with the choice of naming conventions. For the interest of not breaking code, I would prefer to stick with Swift 2 naming (the current naming in ReSwift) conventions. It means we can:

- Release v2.1 with Swift 3 support as it is does not break API compatibility
- Users can move between Swift 2.2 and 3 without any changing any code that relies on ReSwift
- Release v3.0 where we break the API and move to Swift 3 naming conventions
- No documentation changes will be needed as the same usage applies

In this pull request I have:
- Added support for Swift 3
- All tests pass in Xcode 7.3.1 and Xcode 8
- Marked the project as migrated to Swift 3 and Xcode 8 in pbxproj
- Renamed the OS X target and scheme to macOS (this should not break anything as the target module is still called ReSwift)
- Enabled whole module optimisation as suggested by Xcode (again backwards compatible, Xcode 7.3 just ignores the setting)

Thoughts?

### DIFF
--- a/.jazzy.json
+++ b/.jazzy.json
@@ -1,5 +1,5 @@
 {
-    "xcodebuild_arguments": ["-project", "ReSwift.xcodeproj", "-scheme", "ReSwift-OSX"],
+    "xcodebuild_arguments": ["-project", "ReSwift.xcodeproj", "-scheme", "ReSwift-macOS"],
     "swift_version": "2.1.1",
     "author_name": "ReSwift",
     "author_url": "https://github.com/ReSwift/ReSwift",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 #Upcoming Release
 
+**Other**:
+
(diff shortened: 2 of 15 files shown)


Ben-G:
@agentk Thanks a lot for making the effort of migrating this in a backwards compatible way!

Even though the code looks a little painful for now, I think this will be very convenient for many devs that are migrating!

PR #127, opened 2016-07-02
[Tests] Migrate Quick/Nimble testing to XCTest

agentk:
Also along with #126 I think it would be valuable to remove Quick/Nimble in favour of plain old XCTest.

Couple of reasons.
- Quick fails to build for tvOS in Travis CI. This stops binary snapshots from uploading for Carthage releases: https://travis-ci.org/ReSwift/ReSwift/builds/141363703. (Debugging with binary frameworks has improved in Xcode 7.3)
- SPEED: XCTest 1:44mins to build and test for macOS and iOS vs 9:29 with Quick/Nimble for just iOS tests.
- Development: The whole project can be worked and tested on without any dependencies. I know it's not hard to run `carthage build --platform osx,ios`, but with XCTest, there is no step 2.
- Quick and Nimble have diverged platform support into separate Swift 2.2 and 3 branches. We would no longer be able to support Swift 2...3 in one branch if we still relied on Quick/Nimble.

###### Readability:

Instead of:
```
describe("#deinit") {
    it("Deinitializes when no reference is held") {
```
It becomes:
```
func testDeinit() {
    // Deinitializes when no reference is held
```

##### Status:

I am really not sure of the best practice when naming and setting up tests with XCTest. As such, this commit is very rough and can do with a lot of scrutiny and renaming/rewording.

But that being said, it works, compiles, tests and passes with 100% code coverage in Swift 2.2 and 3.


agentk:
This is ready for review now.

Travis is set to check for:
- Xcode 8 - Swift 3 - macOS
- Xcode 7.3 - Swift 2.2 - macOS
- Xcode 7.3 - Swift 2.2 - iOS
- Xcode 7.3 - Swift 2.2 - tvOS

Swift 2.1 compatibility has been removed as the recommended `#if swift(>=2.2)` from Apple is not actually compatible with Swift 2.1 and has been broken for a while without anybody noticing.

Along with the migrating to XCTest I have also cleaned up the '#if swift' style too.

- All internal functions that start with an underscore have the required first parameter in swift 3, but not 2.2. As these are only internal, it makes a good place to start and all occurrences are accounted for so far.
- Where possible, large duplicate chunks of code because of a minor call difference have been moved into a compatibility wrapper.
- Most `#if swift`'s now just wrap function declarations where the @noescape decls have moved position or the function needs to be marked as @discardableResult.

With the next major release, the function declarations can be migrated to Swift 3 style naming and remove even more of the `#if swift` wrappers.

Ben-G:
@agentk Thanks a lot for this migration! I should be able to finish review by end of Wednesday (PST).

kouky:
Wow, the speed difference is worth it alone 💯

ReSwift/CoreTypes/Store.swift
```diff
@@ -91,61 +83,61 @@ public class Store<State: StateType>: StoreType {
...
-                    arguments: getVaList(["nil"]))
-            #else
-                NSException.raise(
-                    "ReSwift:IllegalDispatchFromReducer",
-                    format: "Reducers may not dispatch actions.",
-                    arguments: getVaList(["nil"]))
-            #endif
+        guard !isDispatching else {
+            fatalError("ReSwift:IllegalDispatchFromReducer - Reducers may not dispatch actions.")
         }
```
Ben-G: I unfortunately don't think this change won't work like this. The custom implementation of `fatalError` causes an ambiguity outside of the `ReSwift` module with Swift's `fatalError` (at least in Swift 2.2.1). And forces anyone that imports ReSwift to explicitly call `Swift.fatalError`.

In theory it should be possible to make these test utility methods internal.

In general, even though I like moving away from `NSException`, I'm not sure that the additional code + complexity of the third party library is worth it. I think I would prefer keeping the exception and using `XCTAssertThrowsSpecific` - what do you think?

ReSwiftTests/StoreTests.swift
```diff
@@ -7,239 +7,262 @@
...
 
-// swiftlint:disable function_body_length
-// swiftlint:disable type_body_length
-class StoreSpecs: QuickSpec {
+// swiftlint:disable file_length
+class StoreTests: XCTestCase {
 
-    override func spec() {
+    func testInit() {
+        // Dispatches an Init action when it doesn't receive an initial state
```
Ben-G: What would you think about moving these test comments above the test function? E.g.:

```
// Dispatches an Init action when it doesn't receive an initial state
func testInit() {
   ...
}
```

I file like that might make it more obvious that this is a spec-style comment and a comment related to the source code within the test?

Ben-G:
@agentk Overall this looks really great! Thanks a lot once again!

I had one style comment and I found one regression around the custom implementation of `fatalError`.

Besides that everything looks good to go. I've tested this branch with the counter demo app and everything still worked (with Swift 2.2).

ReSwift/CoreTypes/Store.swift
agentk: I would prefer not to have the additional complexity and overhead to maintain too. I forgot to comeback and check if it could be made private to prevent it leaking 😬.

I'll remove the change. I originally made the change because I was having a lot of trouble getting XCTest to catch the NSException in Swift 3. I think I might have worked it out now though.

ReSwiftTests/StoreTests.swift
agentk: Good call. I'll make them doc comments too.

ReSwift/CoreTypes/Store.swift
agentk: So I remembered what the issue was: `XCTAssertThrowsSpecific` is not available in Swift 2.

We either need to:

- Include the Nimble Objective C shims for catching NSExceptions: https://github.com/Quick/Nimble/blob/master/Sources/Nimble/Adapters/ObjectiveC/NMBExceptionCapture.m
    - Still have to use NSException

- Use a posix signal catcher: http://www.cocoawithlove.com/blog/2016/02/02/partial-functions-part-two-catching-precondition-failures.html
    - Catches NSException or fatalErrors

- Move the refactor the Assertion shim to be internal only.
    - Only catches fatalError

At the moment I have renamed the shimmed `fatalError` to `raiseFatalError` (just to be absolute clear) and marked them all as internal, which still works in testing as long as the framework is `@testable import ReSwift`. I'll commit it after moving the comments.

Happy to look at either of the alternatives or another suggestion.

We COULD make `_defaultDispatch` throw, but that would require refactoring every app that calls `dispatch`.

agentk:
Ok, Ready for review again. Have a think about the fatalError testing options and we will regroup.

Ben-G:
@agentk Going with `raiseFatalError` for now sounds good to me. None of the alternatives seem better. Thanks a lot for looking into this in-depth!

Can we isolate the test case that checks for the `fatalError` into a separate test source file? That way only that source file needs to use `@testable import ReSwift`. I like to keep the scope of testable imports small to ensure we're actually testing the public API.

agentk:
@Ben-G `@testable` refactoring is done. I can't put my finger on it, but the `fatalError` handling still doesn't feel right for some reason.

I picked up a couple of names and descriptions I missed. Others should be able to run with and improve the tests and naming once merged as I have probably been staring at it too long now.

The `@testable` imports are also commented with *why* they are there and what they are required for.

Ben-G:
@agentk I think we should merge this. We can still tweak the `fatalError` handling & test naming in future. What do you think?

agentk:
Good call. Squashed and merged.

Issue #93, opened 2016-03-16
State change with new array items


eschan:
If your state contains an array property, and you append a new item to the array, is there a way to determine which item was added to the state? Is there a way to get a diff between previous state and the new state?

Ben-G:
Hey @eschan,

figuring out which elements have changed is not directly built into ReSwift. However, there are different ways how you can accomplish this.

> Is there a way to get a diff between previous state and the new state?

Yes, however not through a built-in method. You're subscriber can keep a reference to the previous state. That is e.g. how `ReSwiftRouter` works currently: https://github.com/ReSwift/ReSwift-Router/blob/master/ReSwiftRouter/Router.swift#L17

That said, it might be worth adding this directly to the library. Especially when working with UIKit it's often necessary to keep the previous state around for diffing. Another alternative would be building a middleware that takes care of it. Redux has some examples: https://github.com/inakianduaga/redux-state-history.

Lastly, you could use a custom data structure, such as `ObservableArray` (part of the Bond library: https://github.com/SwiftBond/Bond/blob/master/Bond/Core/ObservableArray.swift), to keep track of individual state changes in your array, and add these to your state in some way.

----

There are many possible options, most likely you will want to keep store the previous state in your `Subscriber` and then diff the two arrays (something like Dwifft will help you with that: https://github.com/jflinter/Dwifft)

Hope this helps!

eschan:
thanks for the reply @Ben-G 

I ended up using a combination of RxSwift and AsyncCreator callbacks to manage state transitions and side-effects. 

thanks!

side note: i see that project is now 1.0, congrats!

PR #129, opened 2016-07-04
Keep track of the previous state

gsabran:
Keep track of the previous state to make it possible to react to some specific changes rather than to a given value. For instance, you could do:

```swift
func newState(state: AppState) {
  if let previousState = mainStore.previousState {
    if (state.imageIsLoaded && !previousState.imageIsLoaded) {
      print("image has just loaded")
    }
  }
}
```

### DIFF
--- a/ReSwift/CoreTypes/Store.swift
+++ b/ReSwift/CoreTypes/Store.swift
@@ -21,6 +21,12 @@ public class Store<State: StateType>: StoreType {
 
     // TODO: Setter should not be public; need way for store enhancers to modify appState anyway
 
+    private var _previousState : State?
+    public var previousState: State? {
+        get {
+            return _previousState
+        }
+    }
     /*private (set)*/ public var state: State! {
         didSet {
             subscriptions = subscriptions.filter { $0.subscriber != nil }
@@ -155,7 +161,8 @@ public class Store<State: StateType>: StoreType {
             let newState = reducer._handleAction(action, state: state) as! State
         #endif
         isDispatching = false
-
+        
+        _previousState = state
         state = newState
 
         return action
--- a/ReSwift/CoreTypes/Store.swift
+++ b/ReSwift/CoreTypes/Store.swift
@@ -21,7 +21,7 @@ public class Store<State: StateType>: StoreType {
 
     // TODO: Setter should not be public; need way for store enhancers to modify appState anyway
 
-    private var _previousState : State?
+    private var _previousState: State?
     public var previousState: State? {
         get {
             return _previousState
@@ -161,7 +161,7 @@ public class Store<State: StateType>: StoreType {
             let newState = reducer._handleAction(action, state: state) as! State
         #endif
         isDispatching = false
-        
+
         _previousState = state
         state = newState
 


Ben-G:
@gsabran Thanks a lot for this PR. 🎉 

@ReSwift/reswift I think we hadn't discussed the full tradeoffs of storing the previous state as part of the core framework vs. offering it via some extension, see #93 ?

Sorry for the delay in this @gsabran but it's a somewhat impactful API change that we should discuss thoroughly.

Thanks a lot for the contribution! Let's see what other's have to say about the general direction 😄

agentk:
Nice simple solution @gsabran 👍 

I don't think adding it to the core store is quite the best solution when it's so easy and simple to create a store extension to add the same function. It would also introduce a lot of workarounds required when using replay and jumping to random timeline points. I'd like to keep the PR open though until there are examples in the documentation of how to do this.

The best place to be checking the for previous state is also normally in the ViewModel or even View where you would compare it against the last value displayed and update if they differ. This way the ViewModel can be independent of the Store and also not have to worry about missing a state or double checks on instantiation.

DivineDominion:
Shouldn't pulling this into a subtype work?

In my project, I was considering a Middleware which takes note of the previous state but haven't looked into it. If that'd work, I'd prefer a Middleware as a "core plugin", if you will.

gsabran:
Nice!

@agentk  are you saying that it'd be better to keep reference to the previous state in the view that'd need it, like

```swift
func newState(state: AppState) {
  if let previousState = self.previousState {
    if (state.imageIsLoaded && !previousState.imageIsLoaded) {
      print("image has just loaded")
    }
  }
  self.previousState = state
}
```
?


Also for context, I'm not very experienced with Swift/ReSwift and I'm coming from redux + React.js so any way you would think off that'd offer the same result and that'd integrate better with the ecosystem is perfect.
With React+redux, a state change in redux would trigger a property change in react components that would trigger `componentWillUpdate(newProps, newState)` ([doc](https://facebook.github.io/react/docs/component-specs.html#updating-componentwillupdate)) on the component, and in there you can compare `newProps` to `this.props` (or similarly, after the update, `componentDidUpdate(oldProps, oldState)` is called ([doc](https://facebook.github.io/react/docs/component-specs.html#updating-componentdidupdate)))
So it's more react than redux that makes available the previous state. This also suggests an other approach that would be a protocol similar to `StoreSubscriber` that calls a methods similar to `componentWillUpdate` / `componentDidUpdate` (ie the state comparison would be made available at the component level instead of within redux store). At the end, it'd be closer than what @agentk  suggested with the state thing handled at the component level instead of within the store. Something like

```swift
public protocol StoreSubscriberII: AnyStoreSubscriber {
    #if swift(>=2.2)
    associatedtype StoreSubscriberStateType
    #else
    typealias StoreSubscriberStateType
    #endif

    func stateWillChange(nextState: StoreSubscriberStateType)
}

extension StoreSubscriberII {
    var state: StoreSubscriberStateType?

    public func _newState(state: Any) {
        if let typedState = state as? StoreSubscriberStateType {
            #if swift(>=3)
                stateWillChange(nextState: typedState)
            #else
                stateWillChange(typedState)
            #endif
            self.state = state
        }
    }
}
```

And finally you could do
```swift
class ViewController: UIViewController, StoreSubscriberII {
  func stateWillChange(nextState: nextState) {
    if let state = self.state {
      if (nextState.imageIsLoaded && !state.imageIsLoaded) {
        print("image has just loaded")
      }
    }
  }
}
```

DivineDominion:
The steps you outline sound like the view controller would check for state changes/diffs in `stateWillChange` but not update the view, there; in order to work with the actual data you then have to wait for `newState` nevertheless, do I get this right? If so, your check would fit into `newState` just as well it seems.

I like the idea of performing more atomic changes to the view. I'd put a presenter or similar kind of object between store and view; it'd receive `newState` updates from the store and check for changes. (What @agentk suggested to move into the view model.) Then you can perform changes on view model properties based on the result of the diffing, leaving unchanged values untouched. `didSet` observers or Reactive Cocoa bindings from properties to view components will do the job.

Here's some code with [`bind`/`bindAndFire` from Srdan Rasic](http://rasic.info/bindings-generics-swift-and-mvvm/)'s post to stay framework agnostic.

```
extension Presenter {
  func newState(state: AppState) {
    if let previousState = self.previousState {
      if (state.imageIsLoaded && !previousState.imageIsLoaded) {
        self.viewModel.loadingFinished.value = true
      }
    }
  } 
}

class ViewModel {
  let loadingFinished: Dynamic<Bool> = Dynamic(false)
}

class ViewController: UIViewController {

  @IBOutlet var loadingIndicatior: LoadingIndicatorView!

  var viewModel: ViewModel! {
    didSet {
      viewModel.loadingFinished.bindAndFire { [weak self] in
        self?.loadingIndicatorView.update(loaded: $0)
      }
    }
  }
}
```

Ben-G:
I think we resolved the original question here. If not, please let me know and we can reopen the issue. Thanks!

PR #132, opened 2016-07-20
Fixes Xcode 8 Beta 3 warnings / errors

cuva:
This PR aims to remove warnings and error introduced by the newest Xcode 8 Beta 3.

- Fixes compiler warning for explicit `let` in conditional unwrapping.
- Compiler seem to get confused on which `_newState` method to call, when really there's only one implementation. Differentiating the calls with `SelectedState` argument or `State` seem to do the trick, for now.

This PR will be followed by a `swiftlint` PR that silences a few warnings that currently surface on that branch.

### DIFF
--- a/ReSwift/CoreTypes/Action.swift
+++ b/ReSwift/CoreTypes/Action.swift
@@ -52,7 +52,7 @@ extension StandardAction: Coding {
 
     public init?(dictionary: [String: AnyObject?]) {
         guard let type = dictionary[typeKey] as? String,
-          isTypedAction = dictionary[isTypedActionKey] as? Bool else { return nil }
+          let isTypedAction = dictionary[isTypedActionKey] as? Bool else { return nil }
         self.type = type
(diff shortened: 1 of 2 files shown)


ReSwift/CoreTypes/Action.swift
```diff
@@ -52,7 +52,7 @@ extension StandardAction: Coding {
 
     public init?(dictionary: [String: AnyObject?]) {
         guard let type = dictionary[typeKey] as? String,
-          isTypedAction = dictionary[isTypedActionKey] as? Bool else { return nil }
+          let isTypedAction = dictionary[isTypedActionKey] as? Bool else { return nil }
```
cuva: This warning will be fixed with #132 - it seemed better to address swiftlint related changes in a different chunk.

cuva:
After discussing with @agentk I will be closing this PR as `master` is Swift 3 and will re-submit a PR in regards to the fixes for `SE-0099`.

PR #133, opened 2016-07-20
Silences swiftlint warnings

cuva:
This PR silences the few warnings `swiftlint` emits on `line_length` and `conditional_binding_cascade`.

I understand that these may not be present in master, but should probably be addressed nonetheless.

Is it my understand that the `todo` rule is kept intentionally to make them appear as warnings - which seem fair enough.


cuva:
After discussing with @agentk I will be closing this PR as `master` is Swift 3 and will re-submit a PR in regards to the fixes for `SE-0099`.

PR #139, opened 2016-08-02
Fixed Xcode 8 Beta 4 issues with new Swift 3 changes.

madhavajay:

PR #140, opened 2016-08-03
[Swift 3] Support Swift Build with Latest Toolchain

Ben-G:
Our `Package.swift` file has become a little bit outdated. With the following updates it's possible to use Swift build again.

Also requires a syntactical change that is also part of #139. Depending on when that is merged I'll take the change out of this PR.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,5 +1,10 @@
 import PackageDescription
 
 let package = Package(
-    name: "ReSwift"
+    name: "ReSwift",
+    exclude: [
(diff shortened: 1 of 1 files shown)


Ben-G:
@agentk Would you mind reviewing? 😄

agentk:
My bad. Missed coming back to this one sorry. The test failure looked like a Travis glitch. Reran, and all good to go.

I look firer to being able to use SPM in iOS.

Issue #144, opened 2016-08-28
Why do we lose oldState right away?


buranmert:
In state transitions function (namely `newState()`), don't you think there might be a value of passing `oldState` as a parameter with `newState`?

```swift
func newState(oldState: StoreSubscriberStateType, newState: StoreSubscriberStateType)
```

I personally believe `stateDidChange` would be better but it would mean a huge API change, so I gave my example over `newState` again.

```swift
func stateDidChange(oldState: StoreSubscriberStateType, newState: StoreSubscriberStateType)
```

As an example use case: views can show supplementary subviews in some states (for example loading views) and cleaning-up these subviews in every state transition doesn't feel right to me. If we can also know `oldState`, that kind of cleaning-up tasks can be done within if/switch clauses.

Ben-G:
@buranmert We had a discussion about this in #129. We decided that it's easy enough to implement this functionality on top of `ReSwift` so that we didn't want to add it to the core (some developers that use Signals for store subscriptions wouldn't need it).

Would you mind reading the discussion in #129 and letting us know if it resolves your concerns? Since this issue is coming up frequently we should probably add some sort of reference implementation to the documentation!

buranmert:
@Ben-G sorry, I had a look at issues and PRs before but missed that one.

Yes you are right, that solves the problem and keeps the framework minimal.
But I'm not sure about adding a reference implementation as there can be many ways of doing that, maybe just expressing that was a deliberate structural decision to leave that out should be enough.

I'm closing this issue.

PR #145, opened 2016-08-28
newState is called on subscribers before state changes

buranmert:
**Related issue:** #144 

when newState is called, subscribers have newState value via parameter and also can have oldState value via store

#### Example: 

```swift
func newState(state: FooAppState) {
    if (store.state.FooLoadingState) {
        loadingView.hide()
    }

    switch state {
    case FooBlaBlaState:
    ...
    }
}
```


buranmert:
closing due to unit test failure caused by structural changes

PR #147, opened 2016-09-06
Fixed compile errors for Xcode 8 beta 6

ColinEberhardt:
More breaking changes to the Swift 3 language ...

### DIFF
--- a/ReSwift/CoreTypes/Action.swift
+++ b/ReSwift/CoreTypes/Action.swift
@@ -62,7 +62,8 @@ extension StandardAction: Coding {
     public var dictionaryRepresentation: [String: AnyObject] {
         let payload: AnyObject = self.payload as AnyObject? ?? reSwiftNull as AnyObject
 
-        return [typeKey: type as AnyObject, payloadKey: payload, isTypedActionKey: isTypedAction as AnyObject]
+        return [typeKey: type as AnyObject, payloadKey: payload,
+                isTypedActionKey: isTypedAction as AnyObject]
     }
 }
 
--- a/ReSwift/CoreTypes/Action.swift
+++ b/ReSwift/CoreTypes/Action.swift
@@ -60,9 +60,9 @@ extension StandardAction: Coding {
     }
 
     public var dictionaryRepresentation: [String: AnyObject] {
-        let payload: AnyObject = self.payload ?? reSwiftNull
+        let payload: AnyObject = self.payload as AnyObject? ?? reSwiftNull as AnyObject
 
-        return [typeKey: type, payloadKey: payload, isTypedActionKey: isTypedAction]
+        return [typeKey: type as AnyObject, payloadKey: payload, isTypedActionKey: isTypedAction as AnyObject]
     }
 }
 
--- a/ReSwift/CoreTypes/Store.swift
+++ b/ReSwift/CoreTypes/Store.swift
@@ -15,7 +15,7 @@ import Foundation
  reducers you can combine them by initializng a `MainReducer` with all of your reducers as an
  argument.
  */
-public class Store<State: StateType>: StoreType {
+open class Store<State: StateType>: StoreType {
 
     typealias SubscriptionType = Subscription<State>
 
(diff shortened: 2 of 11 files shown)


ColinEberhardt:
Oops - linting fail!

Let me know what you think of this pull request? If it's something you'd potentially merge, I'll set up linting locally and make sure these issues are fixed.

agentk:
Hi @ColinEberhardt !

Seeing as we have committed to Swift 2.2+ support and have Swift 3 support in master, we do need to keep up with Xcode beta's. Would love to hear your feedback on supporting Swift 3 too.

Our main litmus test is Travis, and it can lag behind the betas a bit at times. I feel bad that you may be fighting Travis on beta support in getting the tests to pass, but I don't want to waste your time on linting if we have other issues with Travis.

Are the test failures just because Travis is lagging or are there actual differences in the way Xcode is being run on Travis?

ColinEberhardt:
> Are the test failures just because Travis is lagging or are there actual differences in the way Xcode is being run on Travis?

The failures are on my side, I didn't run the test suite correctly. I've now go the tests running on Xcode8 beta 6 on Travis:

https://travis-ci.org/ReSwift/ReSwift/builds/158088222

I just need to add the various `#if swift(>=3.0)` conditions to make it work for Swift 2.x

ColinEberhardt:
Hi @agentk - travis is still churning through the build, but I'm confident it will work as I've tested it locally on Xcode 8 and 7.

One big issue is that I've had to remove one of the tests, https://github.com/ReSwift/ReSwift/pull/147/files#diff-af0e4456497f73372ed8f378b67dc21eR27 - the issue here is that the semantics of `public` have changed. In order to extend `Store` within the tests it must be marked as `open`:

https://github.com/apple/swift-evolution/blob/master/proposals/0117-non-public-subclassable-by-default.md

I don't really want to wrap the entire `Store`class in an `#if swift(>=3)` block!

samsonjs:
@ColinEberhardt You should be able to duplicate only the class declaration line itself, something like:

    #if swift(>=3)
        open class Store: ... {
    #else
        public class Store: ... {
    #endif

ColinEberhardt:
@samsonjs unfortunately not. As per [the docs](https://developer.apple.com/library/ios/documentation/Swift/Conceptual/BuildingCocoaApps/InteractingWithCAPIs.html#//apple_ref/doc/uid/TP40014216-CH8-ID34):

> A simple conditional compilation statement takes the following form:

```
#if build configuration
statements
#else
statements
#endif
```

> The statements consist of zero or more valid Swift statements, which can include expressions, statements, and control flow statements.

From your example `open class Store: ... {` is not a valid statement.

samsonjs:
Oh ouch! 😕 That is kind of rough in this case.

ColinEberhardt:
Indeed. Looks like someone else has just raised a pull request for a similar set of changes #148 - as you can see they have had to duplicate the entire class:

https://github.com/tkersey/ReSwift/blob/33f83cf57aba20ea04ce4d4e24c855c4a7f52a66/ReSwift/CoreTypes/Store.swift

Personally I don't think this is worth it just for a single unit test.

Furthermore, I don't think you should be able to subclass the `Store`. With the redux pattern, middleware is the mechanism used to enhance the built in behaviour.

tkersey:
@ColinEberhardt It turns out that using `@testable` allows for subclassing classes defined using `public` when under test. I've updated my PR to use it and made it compatible with Xcode 8 GM.

https://github.com/tkersey/ReSwift/blob/Xcode8-GM/ReSwiftTests/StoreTests.swift#L10

Ben-G:
@ColinEberhardt Thanks a lot for this PR! It seems like there's overlap between this contribution and the one by @tkersey? Have you folks chatted about bringing the changes together in one PR?

> Furthermore, I don't think you should be able to subclass the Store. With the redux pattern, middleware is the mechanism used to enhance the built in behaviour.

Fully agreed! However, allowing subclassing is an important workaround until we've found a better way to extend the store. The power of middleware in Swift is limited, due to static type system. I think we need some good examples of middleware that work without subclassing (e.g. migrating ReSwift-Recorder to middleware) before removing `open` from `Store`.

Would you folks be OK discussing Swift 3 compatibility on [gitter](https://gitter.im/ReSwift/public)? Might be the quickest way to get these changes in.

Thanks so much for the contribution!

ColinEberhardt:
I'm happy for you t go with the PR raised by @tkersey - I think it is more complete. I'll close this one now.

PR #149, opened 2016-09-07
Fix warnings, compiler errors and tests

tkersey:
Updated for Xcode 8 GM.

### DIFF
--- a/ReSwift/CoreTypes/Action.swift
+++ b/ReSwift/CoreTypes/Action.swift
@@ -60,9 +60,11 @@ extension StandardAction: Coding {
     }
 
     public var dictionaryRepresentation: [String: AnyObject] {
-        let payload: AnyObject = self.payload ?? reSwiftNull
+        let payload: AnyObject = self.payload as AnyObject? ?? reSwiftNull as AnyObject
 
-        return [typeKey: type, payloadKey: payload, isTypedActionKey: isTypedAction]
+        return [typeKey: type as AnyObject,
+                payloadKey: payload,
+                isTypedActionKey: isTypedAction as AnyObject]
     }
 }
 
--- a/ReSwift/CoreTypes/Middleware.swift
+++ b/ReSwift/CoreTypes/Middleware.swift
@@ -8,7 +8,12 @@
 
 import Foundation
 
-public typealias DispatchFunction =  (Action) -> Any
+public typealias DispatchFunction = (Action) -> Any
 public typealias GetState = () -> StateType?
+#if swift(>=3)
+public typealias Middleware =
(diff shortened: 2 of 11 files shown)


Ben-G:
@tkersey Thx so much for this contribution! I discussed in #148 that we should figure out what the best strategy for merging Swift 3 changes is. Not entirely sure how much you two PRs overlap - see details on other PR.

Thanks so much for contribution!

ColinEberhardt:
@Ben-G I've had a look at the changes, and besides the odd formatting difference they are fundamentally the same. However, this one uses `@testable`, which allows subclassing / testing of the store, so I think it is a better option.

ColinEberhardt:
@tkersey I have spotted a very odd issue, that I am struggling to understand. I think it might be a compiler bug.

The `AsyncActionCreator` has the following type definition:

```
    public typealias AsyncActionCreator = (
                _ state: State,
                _ store: Store,
                _ actionCreatorCallback: @escaping (ActionCreator) -> Void
        ) -> Void
```

Where the callback is correctly marked as `@escaping`. However, when I compile the library (using Carthage, pointing to this branch), the `ReSwift.h` header file has the following definition:

```
public typealias AsyncActionCreator = (State, ReSwift.Store<State>, (@escaping (State, ReSwift.Store<State>) -> Action?) -> Swift.Void) -> Swift.Void
```

I'll reformat to make that easier to read ...

```
public typealias AsyncActionCreator = (
      State,
      ReSwift.Store<State>,
      (@escaping (State, ReSwift.Store<State>) -> Action?) -> Swift.Void
) -> Swift.Void
```

Notice now that the @escaping attribute has moved inside the brackets, meaning that callback itself is no longer escaping. Which, as this is intended for asynchronous code, is a pretty big issue!

Before I try to create a simpler example (which might not be easy!), and raise an issue against the Swift compiler, I was interested to know whether you observed the same problem?

Ben-G:
@tkersey Thanks a lot for the contribution! 🎉 

We're inviting all contributors to become members of the ReSwift organization - that will make it easier to contribute in future, in case you're interested. If you're not interested, feel free to ignore the invite.

tkersey:
@ColinEberhardt Good catch. That definitely looks like a compiler bugs. There are a couple of issues on the Swift Jira that seem related but nothing specific that I could find. A workaround that I found is to just wrap the entire callback function in parens.

https://github.com/tkersey/ReSwift/blob/4c70fa8922653861dc9316301d3c956664dff421/ReSwift/CoreTypes/Store.swift#L248

ColinEberhardt:
Nice workaround @tkersey - that had me stuck for a while!

PR #151, opened 2016-09-12
Wrap function in parenthesis to properly mark as escaping

tkersey:
There appears to be a compiler bug that applies the escaping to the first argument (which is also a function) and not the entire function.


Ben-G:
🎉  Thanks!

PR #157, opened 2016-10-10
[Swift 3] Remove Swift 2.x Compatibility + Mark Store as Open

Ben-G:
This PR contains two changes:

- **Remove Swift 2.x compatibility** I think it's time to move to Swift 3 only for future releases. For folks that need 2.x older releases should suffice.
- **Mark Store as `open`** I missed the discussion around this in the Swift 3 conversion. I think the store and its members should be marked as `open` until to the next major release. While I definitely think we should use `public` going forward, this is a breaking change. In practice it is breaking my hacky recorder (https://github.com/reswift/reswift-recorder). While this a great way to force me to revisit a better way of implementing recording, I think we should defer breaking the API until to a 3.x release. This way, worst case, the recorder can be migrated to Swift 3 and remain on an rely on a pinned version of ReSwift, as we move on and mark the store as `public`.

Happy to discuss both of these proposals!


Ben-G:
Discussed with @agentk. Merging.

PR #160, opened 2016-10-14
Change AppState counter property from 'var' to 'let'

gorozco58:
Change AppState property from **var** to **let** to avoid state mutability anywhere.

The example would be clearer and better. If you wanna do the State implementation with classes so the properties should be immutable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,12 @@ For a very simple app, that maintains a counter that can be increased and decrea
 
 ```swift
 struct AppState: StateType {
-    var counter: Int = 0
+
+    let counter: Int
+
+    init(counter: Int) {
+        self.counter = counter
+    }
 }
 ```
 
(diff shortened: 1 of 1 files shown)


Ben-G:
@gorozco58 Thanks for your suggestion! Due to the value semantics of Swift structs it is not necessary to mark the property as immutable (from your explanation it looks like you are already aware of that).

We recommend using structs for the app state in order to be able to use `var` properties alongside with value semantics. Having to construct a new value for every state mutation quickly becomes tedious (as the amount of properties grow) and requires additional tools such as lenses for being maintainable.

I'd like to keep this example simple by relying on value semantics of structs; I'm happy to discuss further, but I'll close the issue for now. I'm happy to reopen it in case others also think this would be a good change.

Once again, thanks a lot for the contribution!

Issue #36, opened 2016-01-10
Integration of FRP Frameworks


Ben-G:
Unidirectional Data Flow and FRP play together nicely. We should consider how to provide an integration of FRP frameworks. Currently it's fairly easy to role your own, on top of Swift Flow, as shown in this example: https://gist.github.com/Ben-G/c33486169d68a6f3d12f

We should consider if integrations should become a core part of the framework or not; [related discussion in ReduxKit](https://github.com/ReduxKit/ReduxKit/issues/31).

mosamer:
I'm still exploring the framework but why Rxing the subscriber? not the store itself. Something like that may be:
https://gist.github.com/mosamer/824067f387f66b92c244

I'm exposing an Observable<State> for interested entities (probably views) to subscribe to it. May be add some `slicers` (duals to reducers) to slice out the unneeded chunks of the app state. What do you think?

Ben-G:
@mosamer When I first tested Rx integration I went with this approach as well. However, I was always planning on making `MainStore` a final class, once I came up with another implementation.

Relying on subclassing for adding functionality to `MainStore` is not ideal, since it will always only allow us to use one specific type of store. E.g. currently [Swift Flow Recorder](https://github.com/Swift-Flow/Swift-Flow-Recorder) is also implemented as a `MainStore` subclass; which was a temporary solution that I would like to change moving forward.

What if you want a recording store that also supports exposing an Rx state? That wouldn't be possible if both functionalities are provided as subclasses of `MainStore`. 

This is the main reason why my example uses an approach that can be used on any existing store, without the need for subclassing.

Still in search for the best approach!

Odrakir:
Hi,

I'm trying out ReSwift and mixing it with RxSwift so it fits with my current UI implementation. 

Working from @Ben-G's proposal I came up with this: https://gist.github.com/Odrakir/abf64b0712ec7348e30d960d22960830

In this case I pass a closure to the `state` method so I can map the main state to a substate. Also, as I made states Equatable, I can get updates only when that substate actually changes.

What do you think? Does it make sense?

avdwerff:
@Odrakir did you ever take it further? Your solution seems to nice, is it then that you can mix and match traditional vs rx store subscription? I'm also wondering what you do with the rxMiddleware..

Odrakir:
[...]

PR #171, opened 2016-11-25
Add Observable types to ReSwift

Qata:
Hey everyone,

I've noticed that it's a common pattern that people use FRP in conjunction with ReSwift, so I thought why not make that easy regardless of your FRP implementation.

I'd love feedback regarding naming or anything else that could be improved in this PR.


Qata:
@agentk is looking into the build issues. We can't reproduce them so they're probably a Travis issue.

agentk:
Works on my machine ¯\_(ツ)_/¯

It appears to be linked to the global storing of `Assertions.fatalErrorClosure` not running the second time `expectFatalError` is called when running the tests in Travis.

Ben-G:
@Qata thanks a lot for suggesting this change - looks like really great work!

However, for reasons partly discussed in #36 I don't think this should be part of the core library. One of the nice things about ReSwift (and more generally Redux) is that is a tiny yet extensible library. Examples in #36 show how it's fairy easy to plug in FRP frameworks into the existing library.
Adding this functionality to the core increases the code size significantly, making this a bigger third party dependency to add to other projects. 

I've been using ReSwift and similar architectural patterns with ReactiveCocoa/RxSwift, but I think such integrations should be built on top of the core library instead of built directly in to it.

Happy to discuss more. Thanks again for this suggestion!

Qata:
My intention with adding Observables to ReSwift was to move it away from having many constructs that it currently has which are adding a lot of bloat which Rx implementations already solve, e.g. ActionCreators are simply promises over Actions, subscriptions are just reimplementing Rx subscriptions. Since most people already do use Rx (and to ensure clean code when implemented with ReSwift, *should* be using Rx), my reasoning is that we should move ReSwift in a direction that encourages but does not enforce the usage of third party libraries. The `ObservableProperty` class that I added covers all of the subscription and promise use cases that currently need a lot more code.

Do you think that a pared down and simpler version of ReSwift like that one I'm suggesting (maybe called "ReSwiftRx") could be added to the ReSwift organisation? This would give many benefits to anyone who intends to use Rx with ReSwift without causing any extra bloat to either library. I'd be willing to commit myself to being a core maintainer on such a spin-off project as I believe wholeheartedly in its tenets.

Ben-G:
@Qata I'm definitely open to adding a version of/an add-on for ReSwift that adds Observables out of the box! 

Do you think there's a way to do that in which "ReSwiftRx" or whatever we would call it would have ReSwift as a dependency? So that changes to ReSwift propagate to the other library? If not it's probably not a huge deal as there really isn't a huge amount of code in ReSwift itself.

Qata:
@Ben-G I'd prefer to have it as its own project since there's a lot of code in ReSwift that would become redundant, but we should definitely keep changes in sync between the projects on relevant code and I'd be happy to make sure that happens.

Qata:
I'm hosting it here right now https://github.com/Qata/ReSwiftRx, as soon as you add it to the org I'll just delete my copy. I'll get to work on documentation.

bbqsrc:
ReSwift is already a very small dependency so any code redundancy is a non-issue as far as I'm concerned.

[Moya](https://github.com/moya/moya) supports using Rx or ReactiveCocoa by using submodules basically, and this would be probably the best model to follow, because otherwise you're just making an ReSwift fork, which makes code sharing a pain in the ass, and therefore is less likely to happen at all.

As it stands, I don't see anything in this pull request that actually _removes_ anything from the base RxSwift dependency, so it's already a ripe candidate for a "ReSwift/Rx" submodule.

I think though that the consequences of taking a full side-project path significantly outweighs the benefits.

Qata:
Ideally most of the things added in my PR would've instead done things like replace `Store` with one that only works with observables, remove StoreSubscription entirely, remove ActionCreator since it's just a copy of a `Promise` type, and made reducers use a non-nil StateType to conform with Observables always needing a first value.

If we did separate out a lot of this code into a submodule then it wouldn't restrict the user from using both forms of handling in their app and just bloat the codebase.

My original intention with this PR was to eventually deprecate the similar methods in ReSwift and allow/encourage people to use FRP easily.

Ben-G:
@Qata I've added you to the ReSwift org, so that we can get started on next steps. 

I tend to agree with @bbqsrc:

> As it stands, I don't see anything in this pull request that actually removes anything from the base RxSwift dependency, so it's already a ripe candidate for a "ReSwift/Rx" submodule.

I really think it would be beneficial to build your alternative on the core of `ReSwift`, since it is such a small dependency. Your documentation and example projects could point folks into the right direction of how to use your specific extension.

> If we did separate out a lot of this code into a submodule then it wouldn't restrict the user from using both forms of handling in their app and just bloat the codebase.

I think this can be avoided by providing examples & documentation, what do you think?

Qata:
I think that providing examples and documentation is a good way to go, but providing two ways to do the same thing is my worry, I've seen horrible scenarios arise from similar approaches in other frameworks. 

I'm still trying to flesh out the documentation on my Rx fork, but once it's all done we can review it and see how we want to move things forward regarding any back-porting or spinning out into an extension. I'll make sure to annotate it as experimental in the mean time.

Ben-G:
@Qata got it! I'm definitely open to a totally separate codebase, too. 

Either way, when you add your project, it would be nice to disambiguate clearly in the `README` between this project and `ReSwift/ReSwift` to make sure people are not confused by the two projects.

Qata:
@Ben-G Will do!

john-brunelle:
@Qata Wondering if you were able to get this all fleshed out? Thanks!

Issue #118, opened 2016-05-19
Can we change Reducer from protocol into function type?


soapsign:
Most of the time we need to create a struct to wrap the underlying `handleAction`. If the same functionality can be done with typealias. Should we change to typealias so we don't need to wrap the real  `handleAction` inside a struct.

Ben-G:
Yes, we could potentially do this! The reducer protocol is an artifact from an older ReSwift version. One of the advantages of using a type instead of a free function is that the error message around type mismatches in the store and the root-reducer is slightly better.

However, our docs have been suggesting using free functions for any but the root reducer function; so it might be more consistent to change this.

I don't have a strong opinion either way. I would love to see the change on a branch to see the impact it would have.

@ReSwift/reswift any opinions on this?

DivineDominion:
A reducer type doesn't have much value in my app TableFlip so far compared to a free function.

hartbit:
I also agree that they should be free functions 👍

Ben-G:
I believe @agentk was in favor of this as well. If that's the case I'd be happy to deprecate the `Reducer` protocol with the next release and add support for free functions.

Qata:
I like the idea of deprecating `Reducer`, but what I like even more is turning `Reducer` into a concrete struct that allows mapping and concatenating `Reducer`s with ease. I've implemented this [in ReactiveReSwift](https://github.com/ReSwift/ReactiveReSwift/blob/master/ReactiveReSwift/CoreTypes/Reducer.swift) and I'm very happy with the results. Here's a reimplementation of the GitHub Example code using this new structure.

```swift
let appReducer = Reducer(authenticationReducer, bookmarksReducer, repositoriesReducer)

let authenticationReducer = Reducer<State> { action, state in
    var state = state

    switch action {
    case let action as SetOAuthURL:
        state.authenticationState.oAuthURL = action.oAuthUrl
    case let action as UpdateLoggedInState:
        state.authenticationState.loggedInState = action.loggedInState
    default:
        break
    }

    return state
}

let bookmarksReducer = Reducer<State> { action, state in
    var state = state

    if let action = action as? CreateBookmark {
        let bookmark = (route: action.route, routeSpecificData: action.routeSpecificData)
        state.bookmarks.append(bookmark)
    }

    return state
}

let repositoriesReducer = Reducer<State> { action, state in
    var state = state

    if let action = action as? SetRepostories {
        state.repositories = action.repositories
    }

    return state
}```

[...]

PR #177, opened 2016-12-09
[Reducer] Make Reducer a Function Type

Ben-G:
The renewed discussion on #118 has finally pushed me to actually try out this implementation. Looks great to me. We are already recommending composing reducers through composition as shown in: https://github.com/ReSwift/GitHubBrowserExample/blob/master/SwiftFlowGitHubBrowser/Reducers/AppReducer.swift

This change makes the API more consistent with the recommendation. **Best of all this, change is almost backwards compatible**. Instead of passing a free function you can pass a reference to a method of an instance.

I've done this for all of the unit tests. If we were to merge this I would clean up some of the unit tests (the ones that don't need mocking and thus have stateless reducers) to use free functions instead of types for reducers.


Ben-G:
Ping for discussion: @agentk @DivineDominion @Qata

Qata:
Looks good to me 👍

agentk:
Nice! Once less call needing to force cast too 👌

DivineDominion:
Good job 👏 Solves quite a few cumbersome quirks with the reducer types and looks very straightforward. Going to migrate TableFlip and the demo app as soon as this is integrated in the next release.

kouky:
Nice work on this, stripped away a bunch of unnecessary structs 💯

PR #179, opened 2016-12-14
trying to fix #176

### LINKED ISSUE #176: [Question] Filtered Subscriptions?
if I have substate1, substate2 and uiviews filtered subscribers view1, view2 in one ViewController.
Now when I change substate2, the newState() is being fired for the view2 and for the view1, although substate1 isn't changed. To fix this I confirms the substates to Equatable protocol and when the newState() is fired check if the new substate is equal to the old one and if it's not then change the view properties.

Is there more practical way to do that?
[ReduxTest.zip](https://github.com/ReSwift/ReSwift/files/640703/ReduxTest.zip)

sisomollov:


Ben-G:
Sorry, I'm pretty delayed on feedback, due to traveling. Would you mind providing some more context for this PR? It appears you intend to detect if two states are equal based on a UUID that is maintained by reducers. Is this a pattern that you've used in your projects so far?

#176 Also mentioned that Redux has this feature out of the box. Do you happen to know how it's implemented there? I assume the JSON representation of the state is diffed? But haven't had a chance to check it out yet.

sisomollov:
Hey, sorry for the late reply. My Idea is when we have 2+ substates and 2+ subscriptions in one screen. After user has changed some substate all subscribers get the notification. My solution is to use new subscriber protocol which keep the current substate uuid, if the new state uuid is different, only then call the newState().

All of the current code will work. To get the new functionality have to confirm to this 2 new protocols (EquatableState, EquatableStateStoreSubscriber) and change the uuid in the Reducer when you want to trigger the notification.

mjarvis:
@sisomollov Would you mind taking a look at https://github.com/ReSwift/ReSwift/pull/190 ? It resolves the same issue as this PR, but in a different way. I'm interested to hear why you chose the UUID route vs general Equatable conformance.

sisomollov:
@mjarvis For me the Equatable conformance is a bit annoying, especially if the substate has a lot of properties. That's why I'm using uuid as identifier. 

I think your solution is better than mine, and I still can use my "uuid thing" with one general Equatable state protocol.

Ben-G:
@sisomollov Thanks a lot for this contribution! I believe this issue should be solved with #203. I'll close this PR for now, but feel free to reopen it if you believe it's still relevant!

PR #181, opened 2016-12-16
change back to SWIFT_VERSION = 3.0 for Xcode 8.2

ycfreeman:
should fix #180

PR #182, opened 2016-12-17
Fix swift version for Xcode 8.2

kouky:
The project target fails to build using Xcode 8.2 and when building as a dependency using Carthage.

Fixed by using this
http://stackoverflow.com/questions/41118912/use-legacy-swift-language-version-xcode-8-2


kouky:
closing this, should be fixed by #181

PR #183, opened 2016-12-19
Fixed issue with Xcode complaining about using legacy Swift.

richy486:
Specified directly that this project is using Swift 3.0 for all targets, this was done by turning on and off again the Swift legacy option in the target settings.

This is to help fix the error:

    “Use Legacy Swift Language Version” (SWIFT_VERSION) is required to be configured correctly for targets which use Swift. Use the [Edit > Convert > To Current Swift Syntax…] menu to choose a Swift version or use the Build Settings editor to configure the build setting directly.

When building from Carthage.


Ben-G:
@richy486 Thanks for this PR! I believe the issue is resolved by #181 which was opened a little earlier. Would you mind confirming?

richy486:
@Ben-G Checked out #181 great stuff, no need for this PR, closing

Issue #175, opened 2016-12-08
Middleware is complex and confusing 


Qata:
I think that `Middleware` as it is right now is very complex and confusing to new users. I propose that we deprecate the current method of creating `Middleware` and instead have `Middleware` be a struct that is generic over the type of your state and provides functions such as `map`, `filter` and `flatMap.

[Here is what I've written for ReactiveReSwift](https://github.com/ReSwift/ReactiveReSwift/blob/master/ReactiveReSwift/CoreTypes/Middleware.swift)

Here's an example of how you'd use it:

```swift
let middleware = Middleware<AppState> { getState, dispatch, action in
    print("Received action:")
    return action
}.map { getState, dispatch, action in // `map` is required to not return `nil`
    print(action)
    return action
}.flatMap { action in // `flatMap` can return `nil` to drop the action before it reaches the store
    if action is PopupAction {
        show(action.description)
        return nil
    } else {
        return action
    }
}.filter { action in
    return action is AppAction
}
```

Ben-G:
@Qata there are certainly some interesting ideas in here! Thanks for sharing. One big downside I'm seeing is that the middleware now needs to carry a generic type argument that matches the one of the store. This makes sharing middleware between different projects difficult. This is one of the main reasons we decided not to make the middleware generic.

I'll take a closer look as I go on vacation during the end of the year :)

Ben-G:
@Qata I also renamed the issue so that the title is closer to your suggested changes :)

Qata:
You can still share this `Middleware` between projects with a generic type, all that's required is that the type it's generic over conforms to StateType. 

The current method of creating middleware is not friendly at all and there's no way I'd be able to type it from memory, it was a barrier to entry for me when I first started with ReSwft and I'd like to see it made better.

Qata:
I also want to point out that my implementation using a generic type is incidental, you could easily change the `GetState` function to `() -> StateType` and require casting of the state.

Ben-G:
@Qata that makes sense. By removing the generic type argument we could indeed share Middleware between different projects. I agree that the API looks a lot more intuitive than what we currently have, especially to Swift developers.

IMO I'd like to move forward using this, if it's OK to drop the generic type argument for Middleware sharing. @agentk @DivineDominion what are your thoughts?

@Qata if we use this new API, could you see if the current Middleware in the tests can be implemented in terms of this new API? I strongly assume so, but it would be an important aspect to show people how to migrate from the old to the new API.

[...]

PR #187, opened 2016-12-27
Change dispatch and related functions to return Void

Qata:
As per #175 this is a pull request to remove the `Any` return type from dispatch functions.


Ben-G:
👍 LGTM, thanks!

PR #188, opened 2016-12-27
Change Middleware to a struct with mapping/filtering/side effect capabilities

Qata:
This PR follows on from #187 as discussed in #175

### DIFF
--- a/ReSwift/CoreTypes/Middleware.swift
+++ b/ReSwift/CoreTypes/Middleware.swift
@@ -8,7 +8,7 @@
 
 import Foundation
 
-public typealias DispatchFunction = (Action) -> Any
+public typealias DispatchFunction = (Action) -> Void
 public typealias GetState = () -> StateType?
 public typealias Middleware =
     (DispatchFunction?, @escaping GetState) -> (@escaping DispatchFunction) -> DispatchFunction
--- a/ReSwift/CoreTypes/Middleware.swift
+++ b/ReSwift/CoreTypes/Middleware.swift
@@ -1,14 +1,99 @@
 //
 //  Middleware.swift
-//  ReSwift
+//  FPExamples
 //
-//  Created by Benji Encz on 12/24/15.
-//  Copyright © 2015 Benjamin Encz. All rights reserved.
+//  Created by Charlotte Tortorella on 25/11/16.
+//  Copyright © 2016 Charlotte Tortorella. All rights reserved.
 //
 
-import Foundation
+/**
+ Middleware is a structure that allows you to modify, filter out and dispatch more
+ actions, before the action being handled reaches the store.
+ */
+public struct Middleware {
+    public typealias DispatchFunction = (Action) -> Void
+    public typealias GetState = () -> StateType
 
-public typealias DispatchFunction = (Action) -> Void
-public typealias GetState = () -> StateType?
-public typealias Middleware =
-    (DispatchFunction?, @escaping GetState) -> (@escaping DispatchFunction) -> DispatchFunction
+    internal let transform: (GetState, DispatchFunction, Action) -> [Action]
+
+    /// Create a blank slate Middleware.
+    public init() {
+        self.transform = { [$2] }
+    }
+
+    /**
+     Initialises the middleware with a transformative function.
+     
+     - parameter transform: The function that will be able to modify passed actions.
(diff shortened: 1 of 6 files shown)


ReSwift/CoreTypes/Middleware.swift
```diff
@@ -1,14 +1,99 @@
...
 
-public typealias DispatchFunction = (Action) -> Any
-public typealias GetState = () -> StateType?
-public typealias Middleware =
-    (DispatchFunction?, @escaping GetState) -> (@escaping DispatchFunction) -> DispatchFunction
+    internal let transform: (GetState, DispatchFunction, Action) -> [Action]
+
+    /// Create a blank slate Middleware.
+    public init() {
+        self.transform = { [$2] }
```
Ben-G: For readability I think it would be nice to have an explicit argument name for `action` here. 
E.g.

```swift
public init() {
    // The blank middleware just returns the untransformed action.
    self.transform = { _, _, action in 
           return action 
    }
}
```

While not needed for developers very familiar with ReSwift/FP I think it's a nice way to make it easier for beginners to understand the code/contribute.

ReSwift/CoreTypes/Middleware.swift
```diff
@@ -1,14 +1,99 @@
...
+    public func flatMap(_ transform: @escaping (GetState, Action) -> Action?) -> Middleware {
+        return Middleware { getState, dispatch, action in
+            self.transform(getState, dispatch, action).flatMap {
+                transform(getState, $0)
+            }
+        }
+    }
+
+    /// Drop the Action if `predicate(action) != true`.
+    public func filter(_ predicate: @escaping (GetState, Action) -> Bool) -> Middleware {
```
Ben-G: I think it would be a little clearer to use the argument name the Swift stdlib uses as well: `_ isIncluded:`. I think this is more self-explaining than `predicate`.

ReSwift/CoreTypes/Middleware.swift
```diff
@@ -1,14 +1,99 @@
...
+    public func map(_ transform: @escaping (GetState, Action) -> Action) -> Middleware {
+        return Middleware { getState, dispatch, action in
+            self.transform(getState, dispatch, action).map {
+                transform(getState, $0)
+            }
+        }
+    }
+
+    /// Concatenates the transform function onto the callee's transform.
+    public func increase(_ transform: @escaping (GetState, Action) -> [Action]) -> Middleware {
```
Ben-G: Curious: What are practical uses cases here? The signature suggests this would be used whenever we want to turn an action into multiple separate actions? These actions would then be dispatched separately in the order in which they appear in the array?

Qata: You're right about what the effect of this function is!

Essentially instead of dispatching more values you'd just use `increase` and have multiple values flowing through the middleware. This has the benefit of not needing to dispatch again and go through upper level middleware that has already been passed.

Qata: A possible beneficial use case would be something like an action `PerformLogin` which is expanded to two actions `[PerformLogoutIfNeeded, FetchLoginToken]`.

Ben-G: @Qata Got it. Thant makes sense to me. However, the method name is somewhat unintuitive to me. Also the documentation of `map`, `increase`, etc. is very generic.
Would you mind expanding the documentation a little? And maybe we can come up with a different name for `increase`?

Qata: Yup I'll fix up the documentation and make it more descriptive.

I chose the name `increase` because it's the antonym of `reduce` and the `increase` function does the inverse of the `reduce` function.

I couldn't think of a better name at the time of naming it and I'm still not sure what else I would call it. Do you have any suggestions?

Qata: Oh how about `flatMap`. This mirrors how the standard library has two `flatMap`s, one for arrays and one for optionals.

jondwillis: Perhaps "expand"?

DivineDominion:
@Qata Do you want to continue with this?

Qata:
@DivineDominion nah

PR #189, opened 2017-01-03
Updates documentation and error text to indicate concurrency requirements

langford:
The documentation here is not an academic issue. Not only did I spend several days of time hunting down issues around this, several other iOS programmers on the iOS programming slack immediately came to the same conclusions I originally had that it was permissible to dispatch from a concurrent dispatch queue. 

Many *other* programmers probably don't *know* the global() background queue is concurrent, nor do they currently understand their queue choice is what causes the error as currently written in the codebase.

This PR hopefully adds enough prohibitions to the documentation as well as changes the dispatch exception to a more general, concurrency based exception, thereby letting future programmers understand what is wrong with their code and concurrency strategy.

### DIFF
--- a/Docs/Getting Started Guide.md
+++ b/Docs/Getting Started Guide.md
@@ -193,7 +193,7 @@ func fetchGitHubRepositories(state: State, store: Store<State>) -> Action? {
 
 In this example we're using the `Octokit` library to perform a network request that fetches a users repositories. Within the callback block of the method we dispatch a state update that injects the received repositories into the app state. This will trigger all receivers to be informed about the new state.
 
-Note that the callback block from the network request arrives on a background thread, therefore we're using `dispatch_async(dispatch_get_main_queue())` to perform the state update on the main thread. `ReSwift` will call reducers and subscribers on whatever thread you have dispatched an action from. We recommend to always dispatch from the main thread, but `ReSwift` does not enforce this recommendation.
+Note that the callback block from the network request arrives on a background thread, therefore we're using `dispatch_async(dispatch_get_main_queue())` to perform the state update on the main thread. `ReSwift` will call reducers and subscribers on whatever thread you have dispatched an action from. We recommend to always dispatch from the main thread, but `ReSwift` does not enforce this recommendation. ReSwift *will* enforce that all Dispatches, Store Subscribes and Store Unsubscribes are on the same thread or serial Grand Central Dispatch queue. Therefore the main dispatch queue works, however the global dispatch queue, being concurrent, will fail.
 
 In many cases your asynchronous tasks will consist of two separate steps:
 
--- a/ReSwift/CoreTypes/Store.swift
+++ b/ReSwift/CoreTypes/Store.swift
@@ -107,9 +107,9 @@ open class Store<State: StateType>: StoreType {
     open func _defaultDispatch(action: Action) -> Void {
         guard !isDispatching else {
             raiseFatalError(
-                "ReSwift:ConcurrentMutationError- Concurrent mutation detected. Possible causes: " + 
+                "ReSwift:ConcurrentMutationError- Concurrent mutation detected. Possible causes: " +
                 " A reducer is dispatching an action," +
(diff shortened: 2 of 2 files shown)


Ben-G:
@langford Thanks for this PR! I like the idea of stating the potential causes explicitly. However, there's a small error in the documentation / error message. 

ReSwift does not enforce anything regarding threading (neither for dispatch nor subscription, etc). ReSwift is unaware of threads/GCD or any other concurrency tools. ReSwift only enforces that no action is dispatched while another action is processed. ReSwift *expects* to be used in a non-concurrent environment, which means you should only use it from a single thread / single serial queue.

I think the documentation is already accurate and extending the error message itself should be sufficient. 

The error message should probably read as following (or similar): 

```swift
 "ReSwift:ConcurrencyError- Action has been dispatched while previous action is processed. Possible causes:" +
" A reducer is dispatching an action," +
" ReSwift is used in a concurrent context (e.g. from multiple threads)")
```

I think it's good to add threading as a possible cause to the error message. However, I wouldn't add any more details than that. While some developers might not be very familiar with GCD, ReSwift error messages shouldn't be responsible for explaining the fundamentals.
With the hint about threading, developers should be able to quickly research the problem and find solutions in this repo.

What do you think?

langford:
>ReSwift does not enforce anything regarding threading 
>ReSwift only enforces that no action is dispatched while another action is processed

These are de facto the same thing. Another way to say this in documentation is "ReSwift is not reentrant" or "ReSwift is not threadsafe". It fails intermittently, in a confusing way when used in a threaded environment, so saying "Don't do that" is important. 

> `ReSwift` will call reducers and subscribers on whatever thread you have dispatched an action from. We recommend to always dispatch from the main thread, but `ReSwift` does not enforce this recommendation.

At the very least, this should be rewritten to remove the "Any thread" bit (which makes that sound like an okay way to do things, which it's not), and change it to:

`ReSwift` is designed to have its reducers and subscribers called on a single thread. We recommend to always dispatch from the main thread, but `ReSwift` does not enforce this recommendation. The dispatch handling is not reentrant and will fail if called a second time while executing.

It kinda feels like there should be a hard piece of concurrency logic around a "isDispatching" flag instead of just a flag.

>ReSwift is used in a concurrent context (e.g. from multiple threads)

Are you saying you're not expecting even reads to be in a multithreaded environment?

Ben-G:
> It kinda feels like there should be a hard piece of concurrency logic around a "isDispatching" flag instead of just a flag.

For various reasons this would be complicated / undesirable. We would have to make assumptions about which tools users of ReSwift are using for concurrency (e.g. Threads, GCD, etc.).

Also, ReSwift currently works as a Swift package that doesn't rely on Foundation / GCD and I think we'd like to keep it that way. I still think hinting at potential multi-thread issues in the error message will solve the issue you are rightly concerned about.

> Are you saying you're not expecting even reads to be in a multithreaded environment?

By default Swift is not thread safe and we're not using any other tools to make ReSwift threadsafe (locks, GCD, etc.). This means concurrent reads and writes from multiple threads are not safe. Typically this isn't a concern since it makes sense to only use ReSwift from a single thread.

Ben-G:
@langford Since this PR has been stale for a while I'm going to close it for now. Please feel free to reopen it if you want to continue to work on it!

langford:
Would this be mergeable with:

"ReSwift:ConcurrencyError- Action has been dispatched while previous action is processed. Possible causes:" +
" A reducer is dispatching an action," +
" ReSwift is used in a concurrent context (e.g. from multiple threads)")

The needs discussion tag made me think there were people you wanted looped in, and I was awaiting them.

langford:
(And I cannot reopen, but will happily improve the PR to that so we move to something better than the current error message).

Ben-G:
@langford sorry for the confusion! Re-opened. Yes, would love to merge with updated error message!

langford:
Here we go. It has been updated with that description (slightly changed for clarity).

Ben-G:
Thanks @langford. Merged!

PR #190, opened 2017-01-08
Opt-in Equatable State

### LINKED ISSUE #176: [Question] Filtered Subscriptions?

mjarvis:
This allows users to opt-in to state (or any substate) being Equatable.
This causes `newState` to not be called if the selected state is `Equatable` & unchanged.

I chose to go opt-in since it has the least impact on existing users. If we wanted to make it requred, instead, It is a relatively easy change from here, by making `StateType: Equatable`, and `SelectedState: Equatable` for all `subscribe` functions.

This change is mostly backwards compatible, with no code changes necessary for existing users. The only difference is users who already have been selecting an `Equatable` sub-state will find `newState` not being called as often.

There is a performance implication here since comparisons are made each time the store changes, which can be reduced by using subscription selectors to select the minimum necessary state. (which reduces the equality checks required.)

Fixes #176

TODO:
- [ ] Documentation

### DIFF
--- a/ReSwift/CoreTypes/Store.swift
+++ b/ReSwift/CoreTypes/Store.swift
@@ -17,8 +17,6 @@ import Foundation
  */
 open class Store<State: StateType>: StoreType {
 
-    typealias SubscriptionType = Subscription<State>
-
     // swiftlint:disable todo
     // TODO: Setter should not be public; need way for store enhancers to modify appState anyway
 
@@ -28,7 +26,7 @@ open class Store<State: StateType>: StoreType {
             subscriptions.forEach {
                 // if a selector is available, subselect the relevant state
                 // otherwise pass the entire state to the subscriber
-                $0.subscriber?._newState(state: $0.selector?(state) ?? state)
+                $0.newState(state: state, oldState: oldValue)
             }
         }
     }
@@ -37,7 +35,7 @@ open class Store<State: StateType>: StoreType {
 
     private var reducer: Reducer<State>
 
-    var subscriptions: [SubscriptionType] = []
+    var subscriptions: [AnySubscription] = []
 
     private var isDispatching = false
 
@@ -91,10 +89,43 @@ open class Store<State: StateType>: StoreType {
         where S.StoreSubscriberStateType == SelectedState {
             if !_isNewSubscriber(subscriber: subscriber) { return }
 
-            subscriptions.append(Subscription(subscriber: subscriber, selector: selector))
+            let subscription = Subscription(subscriber: subscriber, selector: selector)
+            subscriptions.append(subscription)
 
             if let state = self.state {
-                subscriber._newState(state: selector?(state) ?? state)
+                subscription.newState(state: state, oldState: nil)
+            }
+    }
+
+    open func subscribe<SelectedState: Equatable, S: StoreSubscriber>
+        (_ subscriber: S, selector: ((State) -> SelectedState)?)
+        where S.StoreSubscriberStateType == SelectedState {
+            if !_isNewSubscriber(subscriber: subscriber) { return }
+
+            let subscription = EquatableSubscription(
+                subscriber: subscriber,
+                selector: selector
+            )
+            subscriptions.append(subscription)
+
+            if let state = self.state {
+                subscription.newState(state: state, oldState: nil)
+            }
+    }
+
+    open func subscribe<SelectedState: Equatable, S: StoreSubscriber>
+        (_ subscriber: S, selector: ((State) -> SelectedState?)?)
+        where S.StoreSubscriberStateType == SelectedState? {
+            if !_isNewSubscriber(subscriber: subscriber) { return }
+
+            let subscription = OptionalEquatableSubscription(
+                subscriber: subscriber,
+                selector: selector
+            )
+            subscriptions.append(subscription)
+
+            if let state = self.state {
+                subscription.newState(state: state, oldState: nil)
             }
     }
 
@@ -154,4 +185,11 @@ open class Store<State: StateType>: StoreType {
         _ store: Store,
         _ actionCreatorCallback: @escaping ((ActionCreator) -> Void)
     ) -> Void
+}
+
+extension Store where State: Equatable {
+    open func subscribe<S: StoreSubscriber>(_ subscriber: S)
+        where State: Equatable, S.StoreSubscriberStateType == State {
+            subscribe(subscriber, selector: nil)
+    }
 }
--- a/ReSwift/CoreTypes/Store.swift
+++ b/ReSwift/CoreTypes/Store.swift
@@ -188,7 +188,7 @@ open class Store<State: StateType>: StoreType {
 
 extension Store where State: EquatableState {
     open func subscribe<S: StoreSubscriber>(_ subscriber: S)
-        where State: EquatableState, S.StoreSubscriberStateType == State {
+        where S.StoreSubscriberStateType == State {
             subscribe(subscriber, selector: nil)
     }
 }
--- a/ReSwift/CoreTypes/Store.swift
+++ b/ReSwift/CoreTypes/Store.swift
@@ -97,7 +97,7 @@ open class Store<State: StateType>: StoreType {
             }
     }
 
-    open func subscribe<SelectedState: Equatable, S: StoreSubscriber>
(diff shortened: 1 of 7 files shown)


Qata:
@Ben-G @agentk @DivineDominion What're your thoughts on this?

mjarvis:
Thought that just came up: It might be a good idea to have a `EquatableState` protocol, instead of using `Equatable` directly, such that users can implement a separate state equitability function if they want it to differ from `==` which may be used elsewhere in their application.
Could leverage `Equatable` to provide a default implementation, as well.
```swift
// ReSwift:
protocol EquatableState {
    func isEqual(to state: Self) -> Bool
}

extension EquatableState where Self: Equatable {
    func isEqual(to state: Self) -> Bool {
        return self == state
    }
}

// Framework user implementation:
struct Default: Equatable, EquatableState {
    let string: String

    static func ==(left: Default, right: Default) -> Bool {
        return left.string == right.string
    }
}

struct Custom: Equatable, EquatableState {
    let string: String

    static func ==(left: Custom, right: Custom) -> Bool {
        return left.string == right.string
    }

    func isEqual(to state: Custom) -> Bool {
        return false
    }
}

let one = Default(string: "test")
let two = Default(string: "test")

one.isEqual(to: two) // true

let three = Custom(string: "test")
let four = Custom(string: "test")

three.isEqual(to: four) // false
```

DivineDominion:
With the changes for `EquatableState`, this is an amazing idea, @mjarvis! I love it.

One thing I ask myself: don't protocol extensions win over concrete type implementations when the variable is of type `EquatableState` in this case? So the checks in the sample code work because `four` is of type `Custom`, but in the ReSwift files, we'd be dealing with `EquatableState`. Did I get this wrong when protocol extensions were introduced?

mjarvis:
@DivineDominion Yes, you're correct, when the variable is of type `EquatableState`, but, in this case, the variable is never that type (and cannot be, since `EquatableState` has `Self` type requirements and therefore cannot be used as a concrete type). ReSwift uses generics in the cases for this type, which follow the behaviour we want.

Example:
```swift
let x: EquatableState = Default(string: "something") // `let x: EquatableState = Default(string: "something")`
```
```swift
func test<T: EquatableState>(left: T, right: T) -> Bool {
    return left.isEqual(to: right)
}

test(left: one, right: two) // true
test(left: three, right: four) // false
```

mjarvis:
So, in implementing `EquatableState`, I've found a major downside -- It means we don't get automatic equatable sub-state selection when selecting something like a `String` or `Int` or `[<T: Equatable>]` etc. We'd have to `extension String: EquatableState {}` etc. That is very finicky, and difficult to maintain for all built-in Equatable types.

DivineDominion:
I don't see this as a big downside because only the apps that use ReSwift will have to decide which built-in `Equatable` types they want to have as substates anyway. As opposed to the status quo, your idea makes support for the "notify on changes only" a one-liner of type-conformance _once_ as opposed to call-site equality checks everywhere. I think that's a good payoff. We only have to communicate this clearly -- or come up with a script that generates `EquatableState`-conformance to all relevant types. Like `Optional<T> where T: Equatable` since `Optional` itself isn't equatable anymore IIRC.

(I, for example, wrap relevant sub-states in custom StateTypes out of convention, so that's never a problem for me and my opinion may be biased.)

mjarvis:
@DivineDominion Yeah, my concern would just be that people would forget to `extension SomeEquatableType: EquatableState {}` for some things, but I suppose thats up to the user to specify which substates should work in that manner. I'm fine with that. I'll fixup this PR to the `protocol EquatableState` style.

Re: `Optional<T> where T: Equatable` sadly Swift 3.0 doesn't support conditional protocol conformance, `Optional: EquatableState where Wrapped: Equatable {}` is a compiler error `extension of type 'Optional' with constraints cannot have an inheritance clause
extension Optional: EquatableState where Wrapped: EquatableState {}`

DivineDominion:
:man_facepalming:  Totally forgot that protocol extension thing. Dang.

Ben-G:
@mjarvis Sorry for being slow on responding to this. Just got back from traveling. Since this is a larger PR it will take a little longer to test it and give feedback. Hope to take some time next week!

Ben-G:
This time for real, blocking off time this weekend to try this PR out. Thanks for patience!

ReSwift/CoreTypes/Store.swift
```diff
@@ -155,3 +186,10 @@ open class Store<State: StateType>: StoreType {
         _ actionCreatorCallback: @escaping ((ActionCreator) -> Void)
     ) -> Void
 }
+
+extension Store where State: Equatable {
+    open func subscribe<S: StoreSubscriber>(_ subscriber: S)
+        where State: Equatable, S.StoreSubscriberStateType == State {
```
DivineDominion: isn't `where State: Equatable` redundant to line 190? Looking at it, I wonder if we should move the method up into the main body with the type requirement specifier directly.

ReSwiftTests/StoreSubscriberTests.swift
```diff
@@ -66,6 +66,28 @@ class StoreSubscriberTests: XCTestCase {
...
+        var state = TestAppState()
+        state.otherTestValue = 3
+        let store = Store(reducer: reducer.handleAction, state: state)
+        let subscriber = TestFilteredNonOptionalSubscriber()
+
+        store.subscribe(subscriber) {
+            $0.otherTestValue
+        }
+
+        XCTAssertEqual(subscriber.receivedValue, 3)
```
DivineDominion: Idea: add `XCTAssertEqual(subscriber.newStateCallCount, 1)` after line 84 as a precondition, too

DivineDominion:
I'm using custom state equatability methods in my project now, too. Switching to a custom equatability check that only accidentally is the same as `==` on the framework level sounds harmless and even a bit helpful.

I wonder what we need to check. Doesn't cause problems on my side, but YMMV...? 

What can break in other people's projects by switching from `SubscriptionType` to `AnySubscription`, for example?

Ben-G:
@mjarvis in preparation of the review, could you help me understand what `OptionalEquatableSubscription` is used for? It seems like it allows subscribing with a substate selector that might return `nil`?

mjarvis:
@Ben-G `OptionalEquatableSubscription` is there because we cannot extend `Optional<T> where T: EquatableState` to conform to `EquatableState`. So yes, it allows subscribing with a substate selector that may return nil. `StoreSubscriberTests:L51` (`func testUnchangedStateSelector()`) Is the test for this use case. (removing the store function that calls to `OptionalEquatableSubscription` causes this test to fail)

ReSwift/CoreTypes/Store.swift
mjarvis: Compiler segfaults if its moved into the main store body.

But yes you're right the second `where State: Equatable` is redundant. I'll remove that.

mjarvis:
@DivineDominion 
> I'm using custom state equatability methods in my project now, too. Switching to a custom equatability check that only accidentally is the same as == on the framework level sounds harmless and even a bit helpful.
> I wonder what we need to check. Doesn't cause problems on my side, but YMMV...?

I'm not sure what you're trying to say here.

> What can break in other people's projects by switching from SubscriptionType to AnySubscription, for example?

`SubscriptionType` was an internal type, so it shouldn't affect users in any way.

Ben-G:
@mjarvis Finally took time today to take a look at this. Really like the idea of adding this to ReSwift; but I think the current implementation is a little more invasive than it needs to be. I prototyped something in #203, which moves the logic of skipping unnecessary updates to the subscription type. As mentioned there, I think this keeps concerns well separated and makes it easier to enhance the subscription type further, in future. 

Would love to hear your thoughts!

mjarvis:
Closing in favour of https://github.com/ReSwift/ReSwift/pull/203

PR #192, opened 2017-01-09
Add subscriber with substate selector to StoreType

mjarvis:
Allows use of StoreType for generic classes which use substate selection, without the creation of a separate protocol.


Qata:
This seems cool. @Ben-G @agentk @DivineDominion what do you think?

DivineDominion:
Doesn't look harmful, but can you show an example? Maybe I'm not too sharp today, but I don't get what you gain from this. Right now this method is bound to the concrete `Store<...>` type instead of `StoreType`. What did you want to do that didn't work before?

Ben-G:
@mjarvis Agree that this makes sense for mocking/testing. Can you rebase, then merge?

PR #193, opened 2017-01-11
Remove convenience init and replace with default value for middleware

Qata:
Since the convenience init only set the Middleware for us, we can replace this with a default value of `[]` and have the Swift compiler auto-generate a convenience init for us.

### DIFF
--- a/ReSwift/CoreTypes/Store.swift
+++ b/ReSwift/CoreTypes/Store.swift
@@ -41,14 +41,10 @@ open class Store<State: StateType>: StoreType {
 
     private var isDispatching = false
 
-    public required convenience init(reducer: @escaping Reducer<State>, state: State?) {
-        self.init(reducer: reducer, state: state, middleware: [])
-    }
-
     public required init(
         reducer: @escaping Reducer<State>,
         state: State?,
-        middleware: [Middleware]
+        middleware: [Middleware] = []
     ) {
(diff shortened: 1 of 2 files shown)


Ben-G:
@Qata Unfortunately this change is not possible, since default initializes are not publicly visible:

> A default initializer has the same access level as the type it initializes, unless that type is defined as public. For a type that is defined as public, the default initializer is considered internal. If you want a public type to be initializable with a no-argument initializer when used in another module, you must explicitly provide a public no-argument initializer yourself as part of the type’s definition.

[Source](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/AccessControl.html#//apple_ref/doc/uid/TP40014097-CH41-ID19)

mjarvis:
@Ben-G this isn't using a default (memberwise) initializer, this is using a default value for one parameter in the initializer. This change should work publicly just fine, since its still using the same initializer, just omitting a parameter.

The source you indicate is referring to this:
```
struct Some {
    let value: String = "test"
    // Swift generates a default initializer `internal init() {}`
}
```

Ben-G:
@mjarvis Right, thanks for pointing this out! I skimmed over "convenience initializer" but didn't take a close look at the actual change.

LGTM, sorry for confusion!

PR #197, opened 2017-01-20
Add DispatchingStoreType protocol

### LINKED ISSUE #174: Protocols without `associatedtype` for store
I'd like to inject the `StoreType` protocol as an ivar to my `ViewModel ==> Model ==> service` classes to let them dispatch the actions.

I'm experiencing the compiler errors similar to those described in the gist below.
https://gist.github.com/dodikk/d5d62002c6cf730c4b059215beeb9203

Having a protocol without the `associatedtype` would let me accomplish that just fine.

mjarvis:
Defines the interface of a dispatching, stateless Store. Useful for dependency injection when your type only cares about dispatching raw actions, and doesn’t need to know about the store’s state.

Fixes #174

### DIFF
--- a/ReSwift.xcodeproj/project.pbxproj
+++ b/ReSwift.xcodeproj/project.pbxproj
@@ -85,6 +85,10 @@
 		81BCBECF1C63167A00AA4F03 /* Subscription.swift in Sources */ = {isa = PBXBuildFile; fileRef = 81BCBECD1C63167A00AA4F03 /* Subscription.swift */; };
 		81BCBED01C63167A00AA4F03 /* Subscription.swift in Sources */ = {isa = PBXBuildFile; fileRef = 81BCBECD1C63167A00AA4F03 /* Subscription.swift */; };
 		81BCBED11C63167A00AA4F03 /* Subscription.swift in Sources */ = {isa = PBXBuildFile; fileRef = 81BCBECD1C63167A00AA4F03 /* Subscription.swift */; };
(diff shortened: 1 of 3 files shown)


Ben-G:
@mjarvis @dodikk After some further thought I'm not sure that this needs to be included in the ReSwift codebase. 

A protocol that only requires the `dispatch` method can be added in the codebase that uses `ReSwift` and the store can be extended to conform to that protocol. This is somewhat idiomatic Swift - apps that consume libraries provide their own protocols for mocking.

mjarvis:
@Ben-G The problem to me is one cannot extend a protocol.
So I have some type that has `<T: StoreType> where T.State: HasSome`, and from it I want to pass the store to something with `var store: DispatchingStoreType?`. I cannot put this protocol in my own codebase since `extension StoreType: DispatchingStoreType {}` is a compiler error.

My options in this case are:
1. Make my type generic with `<T: StoreType>` instead of using `var store: DispatchingStoreType?` (But what if I have some restriction where I can't have this type be generic?)
2. Use a delegate or closure callback to the creator to dispatch actions

DivineDominion:
I'm also _pro_ improving testability in the core library in general. This seems like a first step in that direction. (Currently, I replace `store.dispatchFunction` with a closure that captures the incoming action. When that property changes to non-public or goes away entirely, which is understandable, I'd be screwed.)

Ben-G:
@mjarvis Once again, sorry for being slow on this. 

LGTM! I hadn't considered the use case you described above.

Can you rebase & merge?

PR #200, opened 2017-01-30
Fix build failure when used with SwiftLint 0.16.1

mokagio:
Today I noticed that when building on a machine with SwiftLint 0.16.1 the build fails. This is due to a new SwiftLint rule, shorthand_operator, which has been introduced in version 0.16.1.

This PR updates the code to be complaint with that rule. It simply replaces `let foo = foo + bar` with `let foo += bar`.

An alternative to this option is to disable the rule in the `.swiftlint.yml` config file. I'm happy to close this and submit a new one if that's the preferred style 😄 .

My suggestion is to keep it enabled, as it makes the code more succinct.

### DIFF
--- a/ReSwiftTests/StoreMiddlewareTests.swift
+++ b/ReSwiftTests/StoreMiddlewareTests.swift
@@ -14,7 +14,7 @@ let firstMiddleware: Middleware = { dispatch, getState in
         return { action in
 
             if var action = action as? SetValueStringAction {
-                action.value = action.value + " First Middleware"
+                action.value += " First Middleware"
                 return next(action)
             } else {
                 return next(action)
@@ -28,7 +28,7 @@ let secondMiddleware: Middleware = { dispatch, getState in
         return { action in
 
             if var action = action as? SetValueStringAction {
-                action.value = action.value + " Second Middleware"
+                action.value += " Second Middleware"
                 return next(action)
             } else {
                 return next(action)


Ben-G:
Thanks for this fix @mokagio!

This brings up an interesting point; should we switch to use the CocoaPods version of SwiftLint and pin it to a certain version? While it generally would make sense, it would force developers to use CocoaPods instead of Carthage.

With that in mind, it's probably best to just always support the latest version of SwiftLint?

Either way; thanks a lot for the contribution 🎉! We're inviting all contributors to become members of the org - this will make future contributions easier in case you are interested. If not, feel free to ignore the invite.

mokagio:
No worries @Ben-G I have been using (and plan to use more and more) ReSwift both in my side projects and at work with the team at [iflix](https://iflix.com). It's a pleasure to be able to give back a bit! 😄 .

My two cents:

> This brings up an interesting point; should we switch to use the CocoaPods version of SwiftLint and pin it to a certain version? While it generally would make sense, it would force developers to use CocoaPods instead of Carthage.

For a framework to be _pullable_ only via one dependency manager seems like a limitation to me. I would either not use it, or find a workaround.

> With that in mind, it's probably best to just always support the latest version of SwiftLint?

That's a good question, as keeping it up to date could add some maintenance burden. On the other hand, being up to date with SwiftLint means having a tidier codebase (from a certain point of view at least), so it sound desirable to me.

> Either way; thanks a lot for the contribution 🎉! We're inviting all contributors to become members of the org - this will make future contributions easier in case you are interested. If not, feel free to ignore the invite.

I'd be honoured to be part of the org 💚 .

mjarvis:
> For a framework to be pullable only via one dependency manager seems like a limitation to me. I would either not use it, or find a workaround.

The move to CocoaPods Ben describes wouldn't affect people using the framework in other projects, only for working on the framework itself.

PR #203, opened 2017-02-04
Skip Repeats Implementation

Ben-G:
Thanks a lot to @mjarvis and his work in #190. As part of reviewing his change I noticed some of the aspects that I think could be improved. It would be nice to add a feature that allows skipping state updates that haven't changed a certain substate to the core of ReSwift, as a lot of folks have been asking about that.

However, after some prototyping, I think that the implementation of this should live in the subscription and not in the store. Looking at this prototype, the changes become a lot less invasive that way.

Essentially our subscription type is very, very lightweight reactive component that transforms new values coming out of the store. By keeping the logic in the subscription, we make that reactive component extensible, independent of the rest of ReSwift.  We can add a few operators to it (such as `skipRepeats`) that are going to be used by a majority of developers (however, we should refrain from implementing a full signal library, since enough great libraries exist in that space).

The implementation right now is still pretty hard-wired. I would like to improve this, such that all the transformation logic of new values lives in the subscription and such that various transformations can be composed (as in reactive libraries).

But since I'm almost out of time for this weekend, I wanted to share this working prototype for now.

-----

## To-do list

(by @DivineDominion: Just to end up in the PR overview, I pulled the comments here as a todo)

- [ ] Provide alternative API for creating a subscription that doesn't require using a closure
- [x] Make equatable substates use `skipRepeats` by default
- [ ] Potentially find a new name for `Subscription`/`skipRepeats`, though I haven't heard suggestions on this yet. 
- [ ] document subscription API change:

>         // If the same subscriber is already registered with the store, replace the existing
>         // subscription with the new one.
>         if let index = subscriptions.index(where: { $0.subscriber === subscriber }) {

### DIFF
--- a/ReSwift/CoreTypes/Store.swift
+++ b/ReSwift/CoreTypes/Store.swift
@@ -17,23 +17,17 @@ import Foundation
  */
 open class Store<State: StateType>: StoreType {
 
-    typealias SubscriptionType = Subscription<State>
+    typealias SubscriptionType = SubscriptionBox<State>
 
     // swiftlint:disable todo
     // TODO: Setter should not be public; need way for store enhancers to modify appState anyway
+    // swiftlint:enable todo
 
     /*private (set)*/ public var state: State! {
         didSet {
             subscriptions = subscriptions.filter { $0.subscriber != nil }
             subscriptions.forEach {
-                // if a selector is available, subselect the relevant state
-                // otherwise pass the entire state to the subscriber
-                guard let state = state else {
-                    // State should never be nil after `ReSwiftInit`
-                    preconditionFailure("Unexpectedly found `nil` while unwrapping state")
-                }
-
-                $0.subscriber?._newState(state: $0.selector?(state) ?? state)
+                $0.newValues(oldState: oldValue, newState: state)
             }
         }
     }
@@ -83,19 +77,37 @@ open class Store<State: StateType>: StoreType {
 
     open func subscribe<S: StoreSubscriber>(_ subscriber: S)
         where S.StoreSubscriberStateType == State {
-            subscribe(subscriber, selector: nil)
+            _ = subscribe(subscriber, transform: nil)
     }
 
-    open func subscribe<SelectedState, S: StoreSubscriber>
-        (_ subscriber: S, selector: ((State) -> SelectedState)?)
-        where S.StoreSubscriberStateType == SelectedState {
-            if !_isNewSubscriber(subscriber: subscriber) { return }
+    @discardableResult
+    open func subscribe<SelectedState, S: StoreSubscriber>(
+        _ subscriber: S, transform: ((Subscription<State>) -> Subscription<SelectedState>)?
+    ) where S.StoreSubscriberStateType == SelectedState
+    {
+        // If the same subscriber is already registered with the store, replace the existing
+        // subscription with the new one.
+        if let index = subscriptions.index(where: { $0.subscriber === subscriber }) {
+            subscriptions.remove(at: index)
+        }
 
-            subscriptions.append(Subscription(subscriber: subscriber, selector: selector))
+        // Create a subscription for the new subscriber.
+        let originalSubscription = Subscription<State>()
+        // Call the optional transformation closure. This allows callers to modify
+        // the subscription, e.g. in order to subselect parts of the store's state.
+        let transformedSubscription = transform?(originalSubscription)
 
-            if let state = self.state {
-                subscriber._newState(state: selector?(state) ?? state)
-            }
+        let subscriptionBox = SubscriptionBox(
+            originalSubscription: originalSubscription,
+            transformedSubscription: transformedSubscription,
+            subscriber: subscriber
+        )
+
+        subscriptions.append(subscriptionBox)
+
+        if let state = self.state {
+            originalSubscription.newValues(oldState: nil, newState: state)
+        }
     }
 
     open func unsubscribe(_ subscriber: AnyStoreSubscriber) {
--- a/ReSwift/CoreTypes/Store.swift
+++ b/ReSwift/CoreTypes/Store.swift
@@ -165,3 +165,12 @@ open class Store<State: StateType>: StoreType {
         _ actionCreatorCallback: @escaping ((ActionCreator) -> Void)
     ) -> Void
 }
+
+// MARK: Skip Repeats for Equatable States
+
+extension Store where State: Equatable {
+    open func subscribe<S: StoreSubscriber>(_ subscriber: S)
+        where S.StoreSubscriberStateType == State {
+            _ = subscribe(subscriber, transform: { $0.skipRepeats() })
+    }
+}
--- a/ReSwift/CoreTypes/Subscription.swift
+++ b/ReSwift/CoreTypes/Subscription.swift
@@ -8,7 +8,7 @@
 
 import Foundation
 
-/// A box arounds subscriptions and subscribers.
+/// A box around subscriptions and subscribers.
 ///
 /// Acts as a type-erasing wrapper around a subscription and its transformed subscription.
 /// The transformed subscription has a type argument that matches the selected substate of the
--- a/ReSwift/CoreTypes/Subscription.swift
+++ b/ReSwift/CoreTypes/Subscription.swift
@@ -8,7 +8,121 @@
 
 import Foundation
 
-struct Subscription<State: StateType> {
-    private(set) weak var subscriber: AnyStoreSubscriber?
-    let selector: ((State) -> Any)?
+/// A box arounds subscriptions and subscribers.
+///
+/// Acts as a type-erasing wrapper around a subscription and its transformed subscription.
+/// The transformed subscription has a type argument that matches the selected substate of the
+/// subscriber; however that type cannot be exposed to the store.
+///
+/// The box subscribes either to the original subscription, or if available to the transformed
+/// subscription and passes any values that come through this subscriptions to the subscriber.
+class SubscriptionBox<State> {
+
+    private let originalSubscription: Subscription<State>
+    weak var subscriber: AnyStoreSubscriber?
+
+    init<T>(
+        originalSubscription: Subscription<State>,
+        transformedSubscription: Subscription<T>?,
+        subscriber: AnyStoreSubscriber
+    ) {
+        self.originalSubscription = originalSubscription
+        self.subscriber = subscriber
+
+        // If we received a transformed subscription, we subscribe to that subscription
+        // and forward all new values to the subscriber.
+        if let transformedSubscription = transformedSubscription {
+            transformedSubscription.observe { oldState, newState in
+                self.subscriber?._newState(state: newState as Any)
+            }
+        // If we haven't received a transformed subscription, we forward all values
+        // from the original subscription.
+        } else {
+            originalSubscription.observe { oldState, newState in
+                self.subscriber?._newState(state: newState as Any)
+            }
+        }
+    }
+
+    func newValues(oldState: State, newState: State) {
+        // We pass all new values through the original subscription, which accepts
+        // values of type `<State>`. If present, transformed subscriptions will
+        // receive this update and transform it before passing it on to the subscriber.
+        self.originalSubscription.newValues(oldState: oldState, newState: newState)
+    }
+}
+
+/// Represents a subscription of a subscriber to the store. The subscription determines which new
+/// values from the store are forwarded to the subscriber, and how they are transformed.
+/// The subscription acts as a very-light weight signal/observable that you might know from
+/// reactive programming libraries.
+public class Subscription<State> {
+
+    // MARK: Public Interface
+
+    /// Provides a subscription that selects a substate of the state of the original subscription.
+    /// - parameter selector: A closure that maps a state to a selected substate
+    public func select<Substate>(
+        _ selector: @escaping (State) -> Substate
+        ) -> Subscription<Substate>
+    {
+        return Subscription<Substate> { sink in
+            self.observe { oldState, newState in
+                sink(oldState.map(selector) ?? nil, newState.map(selector) ?? nil)
+            }
+        }
+    }
+
+    /// Provides a subscription that skips certain state updates of the original subscription.
+    /// - parameter isRepeat: A closure that determines whether a given state update is a repeat and
+    /// thus should be skipped and not forwarded to subscribers.
+    public func skipRepeats(_ isRepeat: @escaping (State, State) -> Bool) -> Subscription<State> {
+        return Subscription<State> { sink in
+            self.observe { oldState, newState in
+                switch (oldState, newState) {
+                case let (old?, new?):
+                    if !isRepeat(old, new) {
+                        sink(oldState, newState)
+                    } else {
+                        return
+                    }
+                default:
+                    sink(oldState, newState)
+                }
+            }
+        }
+    }
+
+    // MARK: Internals
+
+    var observer: ((State?, State?) -> Void)? = nil
+
+    init() {}
+
+    /// Initializes a subscription with a sink closure. The closure provides a way to send
+    /// new values over this subscription.
+    private init(sink: @escaping (@escaping (State?, State?) -> Void) -> Void) {
+        // Provide the caller with a closure that will forward all values
+        // to observers of this subscription.
+        sink { old, new in
+            self.newValues(oldState: old, newState: new)
+        }
+    }
+
+    /// Sends new values over this subscription. Observers will be notified of these new values.
+    func newValues(oldState: State?, newState: State?) {
+        self.observer?(oldState, newState)
+    }
+
+    /// A caller can observe new values of this subscription through the provided closure.
+    /// - Note: subscriptions only support a single observer.
+    fileprivate func observe(observer: @escaping (State?, State?) -> Void) {
+        self.observer = observer
+    }
+}
+
+extension Subscription where State: Equatable {
+    public func skipRepeats() -> Subscription<State>{
+        return self.skipRepeats(==)
+    }
 }
--- a/ReSwift/CoreTypes/Subscription.swift
+++ b/ReSwift/CoreTypes/Subscription.swift
@@ -58,11 +58,7 @@ class SubscriptionBox<State> {
 /// reactive programming libraries.
 public class Subscription<State> {
 
-    // MARK: Public Interface
-
-    /// Provides a subscription that selects a substate of the state of the original subscription.
-    /// - parameter selector: A closure that maps a state to a selected substate
-    public func select<Substate>(
+    private func _select<Substate>(
         _ selector: @escaping (State) -> Substate
         ) -> Subscription<Substate>
     {
@@ -71,6 +67,27 @@ public class Subscription<State> {
                 sink(oldState.map(selector) ?? nil, newState.map(selector) ?? nil)
             }
         }
+    }
+
+    // MARK: Public Interface
+
+    /// Provides a subscription that selects a substate of the state of the original subscription.
+    /// - parameter selector: A closure that maps a state to a selected substate
+    public func select<Substate>(
+        _ selector: @escaping (State) -> Substate
+        ) -> Subscription<Substate>
+    {
+        return self._select(selector)
+    }
+
+    /// Provides a subscription that selects a substate of the state of the original subscription.
+    /// If the selected substate is `Equatable` repeated state updates will be skipped.
+    /// - parameter selector: A closure that maps a state to a selected substate
+    public func select<Substate: Equatable>(
+        _ selector: @escaping (State) -> Substate
+        ) -> Subscription<Substate>
+    {
+        return self._select(selector).skipRepeats()
     }
 
     /// Provides a subscription that skips certain state updates of the original subscription.
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -4,3 +4,4 @@ excluded: # paths to ignore during linting. overridden by `included`.
 
 disabled_rules: # rule identifiers to exclude from running
   - force_cast
+  - opening_brace
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,44 @@
-#Upcoming Release
+#4.0.0
+
+*Work in Progress*
+
+**Breaking API Changes:**
+
+- Introduced a new Subscription API (#203) - @Ben-G
+
+  - The subscription API provides basic operators, such as `skipRepeats` (skip calls to `newState` unless state value changed) and `select` (sub-select a state).
+
+  - This is a breaking API change that requires migrating existing subscriptions that sub-select a portion of a store's state:
+
+    - Subselecting state in 3.0.0:
+
+      ```swift
+      store.subscribe(subscriber) { ($0.testValue, $0.otherState?.name) }
+      ```
+    - Subselecting state in 4.0.0:
+
+      ```swift
+      store.subscribe(subscriber) {
+        $0.select {
+          ($0.testValue, $0.otherState?.name)
+        }
+      }
+      ```
+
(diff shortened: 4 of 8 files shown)


Dimillian:
So you defer the old/new states comparaison to the end user implementation. I like it! More discrete than forcing Equatable for skipping. And I can use my already Equatable state in the shouldSki. As it is, I love it!

Ben-G:
Update:

I've improved the implementation along the lines discussed above. The subscription would now act as a lightweight signal. Subscribers can transform that subscription when they subscribe to the store. Currently we provide `select` and `skipRepeats` transformations. 

The store sends all new values through these subscriptions; the subscriptions forward the transformed values to their subscribers.

I'm pretty happy with the current implementation and would love to hear your thoughts.

As mentioned earlier we should keep the transforms on subscriptions limited since I want to avoid implementing a reactive library as part of ReSwift. If you need the full power of reactive programming and its operators, it's easy enough to use a separate library and integrate it with ReSwift. That said, I like the idea of providing the most necessary transforms out of the box.

Dimillian:
I use your branch in a fairly complex app! I've made all my state conform to Equatable, I've implemented the new select and skipRepeats() subscribers, it works flawlessly!

ReSwift/CoreTypes/Store.swift
```diff
@@ -17,18 +17,17 @@ import Foundation
...
+    // swiftlint:enable todo
 
     /*private (set)*/ public var state: State! {
         didSet {
             subscriptions = subscriptions.filter { $0.subscriber != nil }
             subscriptions.forEach {
-                // if a selector is available, subselect the relevant state
-                // otherwise pass the entire state to the subscriber
-                $0.subscriber?._newState(state: $0.selector?(state) ?? state)
+                $0.newValues(oldState: oldValue, newState: state)
```
DivineDominion: This alone is a big improvement already!

ReSwift/CoreTypes/Store.swift
```diff
@@ -78,19 +77,37 @@ open class Store<State: StateType>: StoreType {
...
-        where S.StoreSubscriberStateType == SelectedState {
-            if !_isNewSubscriber(subscriber: subscriber) { return }
+    @discardableResult
+    open func subscribe<SelectedState, S: StoreSubscriber>(
+        _ subscriber: S, transform: ((Subscription<State>) -> Subscription<SelectedState>)?
+    ) where S.StoreSubscriberStateType == SelectedState
+    {
+        // If the same subscriber is already registered with the store, replace the existing
+        // subscription with the new one.
+        if let index = subscriptions.index(where: { $0.subscriber === subscriber }) {
```
DivineDominion: Oooh, that's nice. It's also an API change that we should note as a TODO for the docs.

DivineDominion:
I really like what happens here. The new types appear less hacky because they have no underscored methods, which is a nice change. 

Reading the new code,  it looks like you've been doing even more reactive programming in the past months :)

The potential of the new `Subscription` is huge. But I'm torn: one could extend the functionality to add `filter` and whatnot, but then again there are reactive frameworks of various flavors to use for such tasks. Is the new `Subscription` type open enough to be easily converted into a `RxSwift.Observable`, for example? (It seems to be trivial to extend it with an `asObservable()` method.) 

This in turn makes me wonder: should this really become part of the core, or is this already a "reactive" extension? (Folks said React.js doesn't propagate idempotent updates. And why should they? That would only expose the internal receiving of an action as an output signal without value. Why should ReSwift do that? @mjarvis's approach is closer to making sense here.)

I'm fine with this approach of lifting state event filtering into a new type because of the potentials. But it changes directions a bit. I wonder if @mjarvis finds any downsides when compared to his approach?

Dimillian:
IMO as it is, only implementing it for Equatable should be a part of this library, I don't want to add anything else to it, and you can already create a syntax somewhat like the @connect

I use it like that 

```Swift
extension Subscription {

    public func connectAll() -> Subscription {
        return self.skipRepeats({ (_, _) -> Bool in
            return false
        })
    }

    public func connectUsersAndLibrary() -> Subscription {
        return self.skipRepeats({ (old, new) -> Bool in
            if let old = old as? AppState, let new = new as? AppState {
                return old.usersState == new.usersState &&
                    old.libraryState == new.libraryState
            }
            return false
        })
    }
}

```

So I can connect and observe only the relevant part of my state when I need it.

mjarvis:
This seems to tick most of the same boxes as what I was intending with my EquatableState work.

What I like:
1. This is a lot more clear in terms of documenting & readability on what will skip repeats, theres no "invisible" stuff happening here.

My concerns:
1. This is a breaking change, as opposed to my work which is backwards-compatible. (minor concern, just means a major version bump, and a bit of work for devs, and might be fixable)
2. The API seems odd, in that the select / skip is in a closure.
```swift
store.subscribe(subscriber) {
      $0.select {
            ($0.testValue, $0.otherState?.name)
       }.skipRepeats()
}

// I feel like I should be able to do this:

store.subscribe(subscriber)
    .select { ($0.testValue, $0.otherState?.name) }
    .skipRepeats()
```

If we can achieve my second example there, we could likely add backwards-compatibility since `select` is function on the subscription instead of the selection closure being part of the `subscribe` function. (and deprecate the old style.)

> @DivineDominion: This in turn makes me wonder: should this really become part of the core, or is this already a "reactive" extension? (Folks said React.js doesn't propagate idempotent updates. And why should they? That would only expose the internal receiving of an action as an output signal without value. Why should ReSwift do that? @mjarvis's approach is closer to making sense here.)

I do wonder this myself, as well. Part of me doesn't think the user should know, and that skipping by default on `Equatable` is desirable for the library. I think if this style is the desirable change (IE: Manual skipping via `skipRepeats()` or similar), is this functionality better suited to allowing a user to extend the subscription with an observable of some kind to give the skipping? Should we support closure based callbacks then, to have that, as well (instead of `func newState`)?

Ben-G:
@DivineDominion I feel like adding minimal support to the core makes sense. As this is something essentially everyone working with ReSwift will need. And it would be nice to provide a solution that works out of the box.

@mjarvis unfortunately I don't see a way to get rid of the closure; as the operators, such as `map` need to return a new type with a different generic type argument and cannot simply mutate the existing type. I definitely agree that your suggested syntax would be preferable, if achievable.

Overall I feel like while this is a small breaking API change, and the API is not entirely perfect, it is still easy to demonstrate through an example and thus should be easy to use.

> Should we support closure based callbacks then, to have that, as well (instead of func newState)?

I think the advantage of the current approach is that we don't need to deal with retain cycles in closures.

----

I generally would be in favor in moving forward with this API as I think it solves an immediate need while keeping the extensions to subscriptions isolated such that they can easily be changed in future. But I think we should take some more time to discuss it.

mjarvis:
Just thinking out loud here, with a couple of alternate API designs (since the nested closure just feels so weird to me)

```swift
// I don't think this one works because it breaks the subscriber + selected state generic relationship, but it sure looks nice :(
// Allows extension of subscription
let subscription = Subscription(subscriber: subscriber)
    .select { $0.testValue }
    .skipRepeats()
store.subscribe(subscription)
```

```swift
// keeps the subscriber + generic selectedstate relationship because they're in a single function call
// Allows extension of subscription
let subscription = Subscription()
    .select { $0.testValue }
    .skipRepeats()
store.subscribe(subscriber, with: subscription)

// Also can be written in alternate fashions such as:
store.subscribe(subscriber,
                with: Subscription()
                    .select { $0.testValue }
                    .skipRepeats()
)
// can also have default of `with subscription: Subscription = Subscription()`
// backwards compatibility via old API creating `Subscription` object internally using the selection closure
)
```

```swift
// most similar to existing API, backwards compatible via `skipRepeats: Bool = false` argument
// Doesn't have the nice, public, extendable Subscription type
store.subscribe(subscriber, skipRepeats: true) {
    $0.testValue
}
```

DivineDominion:
@mjarvis I think (2) is superior to (1) because a subscription depending on the subscriber reads weird -- though the brevity on its own sure is nice.

So when you have this:

```
let subscription = Subscription()
    .select { $0.testValue }
    .skipRepeats()
store.subscribe(subscriber, with: subscription)
```

... you essentially pull the "Builder" out and prepare it in advance. This should be 1:1 convertible to @Ben-G's closure where `$0` simply is the Subscription builder.

Ben-G:
@DivineDominion @mjarvis Would anyone be opposed to support both APIs? That's essentially just syntactic sugar. I personally still prefer the closure based approach as it requires less boilerplate code and no additional local variable. But I get your point about why the API might feel unintuitive to some folks. Don't see much downside in supporting both?

ReSwift/CoreTypes/Store.swift
Ben-G: 👍

DivineDominion:
Support for both 

```
let subscription = Subscription()
    .select { $0.testValue }
    .skipRepeats()
store.subscribe(subscriber, with: subscription)
```

and 

```
store.subscribe(subscriber) {
      $0.select {
            ($0.testValue, $0.otherState?.name)
       }.skipRepeats()
}
```

sounds good to me since they are indeed only superficially different 👍

I'm still concerned about what kind of filters `Subscription` _should_ support without duplicating reactive framework features. But I cannot seem to solve that problem analytically, so we might as well move on :)

Ben-G:
> I'm still concerned about what kind of filters Subscription should support without duplicating reactive framework features. But I cannot seem to solve that problem analytically, so we might as well move on :)

I think we'll see as we go along; but most likely the current set should be the maximum. However, when working with UIKit essentially everyone using ReSwift will need a way to select substates and skip updates on these - this is required to keep the UI updates minimal.

With Redux and React this isn't necessary because minimizing UI re-rendering is conveniently handled by React itself.

For this reason I like the idea of providing the core subscription features I'm proposing because they will make ReSwift usable without requiring the use of a Reactive library. That said, there are still a bunch of uses cases in which advanced users might prefer using a Reactive library over the built in subscription.

danielmartinprieto:
Hi! Just wanted to point out some things: the official way of binding the state changes of a Redux store to a component in React is through a library called React Redux, via its connect function.

The default behavior of this function makes some comparisons (with default functions) before letting the changes propagate to the React component (aka before React does its magic with the virtual DOM). For example, if the previous state is the same (`===`) as the current one, it doesn't trigger the update. If the result of the selector with the current state is the same as the result of the selector with the previous state (the famous `mapStateToProps`), it doesn't trigger it either. That's why it's recommended a lightweight implementation of this function, using memoized selectors.

You could override this default behavior with custom functions that do more than the `===` check (objects in JavaScript are reference types), and therefore make it be even more performant.

In ReSwift, this connection is called "subscription" and this makes it seem like a reactive library, but if you think about it, just removing the skipRepeats (another "reactive" term) and make it the default behavior would be totally fine. We could think later about ways of customize this implementation from the caller, without still having any relation with the reactive world.

DivineDominion:
Good points. Making `skipRepeats` the default makes sense, but how would you turn it off?

React's [`setState`](https://facebook.github.io/react/docs/react-component.html#setstate) triggers a re-rendering, too, unless you return `false` in [`shouldComponentUpdate`](https://facebook.github.io/react/docs/react-component.html#shouldcomponentupdate). 

A React.js component is far more powerful than a ReSwift subscriber, even in this regard. It doesn't make sense to let a "subscriber" have a "shouldSubscriberUpdate" delegate callback. Subscribers are made to receive events. We have torn `React.Component` in two: a state change subscriber and a controlling component, but providing just the former.

_Aside: Maybe this is an indicator we're beginning to miss React-specific features in this Redux-inspired library because there's nothing like it? I haven't worked with React.js myself and cannot say much about how it is different._

Now the interesting question is: should the subscriber receive `newState` even when `state` is not _new_ at all? Doesn't make much sense in practice except when you want to force-redraw views for some reason. Rather sounds like a bug, getting a "new state" update with old state.

Ben's approach is backwards-compatible and introduces a nice add-on feature that a lot of people will have missed. It's not an invasive change, so I think we should roll it out and experiment with it in the field.

_But_ it seems there's need for discussion about the design principles for an upcoming major release.

danielmartinprieto:
I agree it doesn't make sense to receive an update when there is no change in the state. If you want to update the subscriber because other reasons, you'll find a way to do it independently of ReSwift, and if it's because a change of the state, an update will be trigger. I also see no point in let the subscriber turn it off.

Regarding React/Redux:
- Redux always propagates the changes.
- The React Redux binding function `connect` (more or less our `subscribe` function) provides the ability to stop the Redux state changes propagation with some default functions you can override passing them to `connect`. It's true that here we're already in between Redux and React, so it's hard to tell if adopting something like this it's a ReSwift responsibility.
- Inside your component in React, you still can implement the `shouldUpdateProps` function, return false, and don't let your component update, even if the Redux state changed.

So based on all this, I'd make the `skipRepeats` the default and without being exposed to the caller, and maybe in the future think about a way of customising this behaviour based on a function you pass upon subscription, the same way `connect` does.

DivineDominion:
Problem is, we can't :) ... Or at least we couldn't, before Ben's `SubscriptionBox` came along 🤔

```
extension Subscription where State: Equatable {
    public func skipRepeats() -> Subscription<State>{
        return self.skipRepeats(==)
    }
}
```

That doesn't work when `State` is not `Equatable`. But that's not the only option left right now because Ben's PR lifts making decisions about a state change into a new type with new powers.

Maybe it works if we don't add a hard constraint on `StateType` being `Equatable` but instead push the decision-making to the user:

```swift
    @discardableResult
    open func subscribe<SelectedState, S: StoreSubscriber>(
        _ subscriber: S, 
        skipRepeats stateIsEqual: (SelectedState?, SelectedState?) -> Bool,
        transform: ((Subscription<State>) -> Subscription<SelectedState>)?
    ) where S.StoreSubscriberStateType == SelectedState
    {
        // ... mostly like Ben's, but:
        
        let subscriptionBox = SubscriptionBox(
            originalSubscription: originalSubscription,
            transformedSubscription: transformedSubscription,
            subscriber: subscriber
        ).skipRepeats(stateIsEqual) // <- ⚠️
        
        // ...
    }
    
    // This is the supposed default:
    @discardableResult
    open func subscribe<SelectedState, S: StoreSubscriber>(
        _ subscriber: S, 
        transform: ((Subscription<State>) -> Subscription<SelectedState>)?,
    ) where S.StoreSubscriberStateType == SelectedState, 
            S.StoreSubscriberStateType: Equatable
    {
        // ... can infer `skipRepeats`
    }
```

It's simple to add a default `skipRepeats = { _, _ in return false }` to maintain the current behavior.

Ben-G:
@DivineDominion Yep, I like the idea of making it the default where possible. And requiring the user to add `skipRepeats` closure for substates that are not equatable. Also happy to rename `skipRepeats` if we find something that's more intuitive. I borrowed the naming from ReactiveCocoa since I think it describes the behavior pretty well.

mjarvis:
👍 to the great iteration happening here.

I think I'd also prefer skipping repeats to be the default behaviour when equatable is available.

Ben-G:
I'd like to put some finishing touches onto this PR this week. Let's make sure I've got all concerns covered:

- [ ] Provide alternative API for creating a subscription that doesn't require using a closure
- [ ] Make equatable substates use `skipRepeats` by default
- [ ] Potentially find a new name for `Subscription`/`skipRepeats`, though I haven't heard suggestions on this yet.

DivineDominion:
Looks good! Need a hand on this?

Ben-G:
@DivineDominion Thanks, but all these changes are pretty small so I should be able to get them done by myself!

ReSwift/CoreTypes/Subscription.swift
```diff
@@ -8,7 +8,138 @@
...
+
+    /// Provides a subscription that selects a substate of the state of the original subscription.
+    /// If the selected substate is `Equatable` repeated state updates will be skipped.
+    /// - parameter selector: A closure that maps a state to a selected substate
+    public func select<Substate: Equatable>(
+        _ selector: @escaping (State) -> Substate
+        ) -> Subscription<Substate>
+    {
+        return self._select(selector).skipRepeats()
+    }
```
Ben-G: @mjarvis @DivineDominion This is where the default implementation of `skipRepeats` for `Equatable` substates lives.

Ben-G:
@DivineDominion @mjarvis Sorry for the delay on this. But I'm in the last two weeks of major project at work 😄 

I've added a commit that defaults to skipping repeats for any selected substates that are `Equatable`. We could also consider adding it as a default for subscriptions that don't use `select`, but I think that won't be too useful in practice.

When it comes to the alternative subscription API I've got "bad" news. I didn't think through that case earlier; but I noticed that I need access to both the original `Subscription<State>` and the transformed `Subscription<SelectedState>`. That's required because the original subscription is the only one that has the correct type arguments that allows me to pass `<State>` values in. So I'm not sure how (if at all) an API like this would be possible:

```swift
let subscription = Subscription()
    .select { $0.testValue }
    .skipRepeats()
store.subscribe(subscriber, with: subscription)
```
Right now I would need to require the caller to pass in the original subscription as well:

```swift
let subscription = Subscription()
let transformedSubscription = subscription
    .select { $0.testValue }
    .skipRepeats()
store.subscribe(subscriber, with: transformedSubscription, original: subscription)
```

In my personal opinion though the closure based API is fine and fairly easy to understand and thus I don't think it's worth putting a lot of work into an alternative API. But I know that @mjarvis had a different opinion on this. So if anyone else has ideas of how to bring this forward I'd love to hear them. I'd like to see this functionality go into ReSwift and a new release soon (independent of exact API).

However, I'll still be occupied by a work project for the next 1.5 weeks.

ReSwift/CoreTypes/Subscription.swift
```diff
@@ -8,7 +8,138 @@
 
 import Foundation
 
-struct Subscription<State: StateType> {
-    private(set) weak var subscriber: AnyStoreSubscriber? = nil
-    let selector: ((State) -> Any)?
+/// A box arounds subscriptions and subscribers.
```
DivineDominion: Found a typo :) "arounds/around"

DivineDominion:
Just to end up in the PR overview, I pulled my comment here as a todo:

- [ ] document subscription API change

>         // If the same subscriber is already registered with the store, replace the existing
>         // subscription with the new one.
>         if let index = subscriptions.index(where: { $0.subscriber === subscriber }) {

ReSwift/CoreTypes/Store.swift
```diff
@@ -83,19 +77,37 @@ open class Store<State: StateType>: StoreType {
...
+            subscriptions.remove(at: index)
+        }
 
-            subscriptions.append(Subscription(subscriber: subscriber, selector: selector))
+        // Create a subscription for the new subscriber.
+        let originalSubscription = Subscription<State>()
+        // Call the optional transformation closure. This allows callers to modify
+        // the subscription, e.g. in order to subselect parts of the store's state.
+        let transformedSubscription = transform?(originalSubscription)
```
DivineDominion: This line in conjunction with `SubscriptionBox.newValues` paint a complete picture -- but starting to read at `newValues`, the comments tell me what is supposed to happen, but I couldn't see why `transformedSubscription` would be affected.

What do you think about passing the optional transformation closure to SubscriptionBox, too? Then assembling and using the transformed subscription happen in the same file.

mjarvis:
> I've added a commit that defaults to skipping repeats for any selected substates that are Equatable. We could also consider adding it as a default for subscriptions that don't use select, but I think that won't be too useful in practice.

I've actually been switching away from using select, and instead been using protocols that define chunks of state.
I know I could manually provide the skips implementation for this but it would be convenient to have it automatically if I make my `SomeState` protocol Equatable

```swift
protocol SomeState {
    var foo: Foo { get }
    var bar: Bar { get }
}

class Subscriber<S: StoreType>: StoreSubscriber where S.State: SomeState {
    init(store: S) {
        store.subscribe(self)
    }
    
    func newState(state: S.State) {
        // can access state.foo, state.bar
    }
}
```

> That's required because the original subscription is the only one that has the correct type arguments that allows me to pass <State> values in. So I'm not sure how (if at all) an API like this would be possible:

I haven't had time to look too closely, but would manually specifying the generic work for this?
```swift
let subscription = Subscription<State>() // original state type
    .select { $0.testValue }
    .skipRepeats()
```

I suppose this might require a select to be `func select<SelectedState>(_ transform: (State) -> (SelectedState)) -> Subscription<State, SelectedState>`
Might add a factory function to provide a helper with defaults, IE `static func for<S: StoreType>(_ store: S) -> Subscription<S.State, S.State>()`, used like `let subscription = Subscription.for(self.store)` so people don't have to do `let subscription = Subscription<AppState, AppState>`


>  Potentially find a new name for Subscription/skipRepeats, though I haven't heard suggestions on this yet.

In some work I've done recently I've found I need a way to skip newState calls for different reasons, not just for a "repeat". For instance, I need to skip calling newState while some network requests are loading. Perhaps the naming should just be `skip(when: (oldState, newState) -> Bool)`. Example usage:

```swift
subscription.skip(when: { [unowned self] _, state in
    return !state.completedNetworkRequests.contains(self.issuedRequests)
})
```

DivineDominion:
> I've actually been switching away from using select, and instead been using protocols that define chunks of state.

That's weird and interesting at the same time :) But I agree that if you want to skipRepeats for the whole app state it's weird to have `subscribe(subscriber) { $0.select { $0 } }` work thanks to the convenience version of `select<S>(...) where S : Equatable` -- but no way to get this for the main state directly.

Adding yet another method can work, but it might not be a good fit to the existing solutions:

```
store.subscribe(subscriber, skipRepeats: true)
```

> I suppose this might require a select to be func select<SelectedState>(_ transform: (State) -> (SelectedState)) -> Subscription<State, SelectedState>

That's the thing -- we'd have to carry the history around as part of the type through each transformation. What if you have two transformations of any sort? `Subscription<Subscription<OriginalState, FirstSelect>, SecondSelect>`? Couldn't come up with anything clever to get rid of this last weekend :/ Instead I wondered if getting rid of the closure is really a pressing issue. I don't find it too bad as it is right now.

mjarvis:
> I wondered if getting rid of the closure is really a pressing issue. I don't find it too bad as it is right now.

I don't think its a pressing issue, but something I'd love to have if we can figure it out.

> Adding yet another method can work, but it might not be a good fit to the existing solutions:

Just so I'm understanding things correctly:

```swift
store.subscribe(subscriber) {
    $0.select { $0.testValue } // If `testValue` is `Equatable` this infers `skipsRepeats`?
}
```

If the above is true, we should be able to do that without the block by using a function similar to what I did in my equatable PR:
```swift
extension Store where State: EquatableState {
    open func subscribe<S: StoreSubscriber>(_ subscriber: S)
        where S.StoreSubscriberStateType == State {
        _ = subscribe(subscriber, transform: { $0.skipRepeats() })
    }
}
```

DivineDominion:
I think that's the way it works, yes, so the extension sounds like a good idea.

Ben-G:
Trying to now close the loop on this 😃 

@mjarvis Your extension sounds like good idea! Can you remind me why you used `EquatableState` instead of `Equatable`? And what did the type definition for `EquatableState` look like?

mjarvis:
> @mjarvis Your extension sounds like good idea! Can you remind me why you used EquatableState instead of Equatable? And what did the type definition for EquatableState look like?

`EquatableState` was used to give people an option to provide a separate implementation for the skipping check. Its not necessary in your implementation because if someone wants a separate implementation they can implement the `skips` closure, instead of using the default implementation. (You can use `Equatable` in my example instead)

Ben-G:
@mjarvis @DivineDominion Thanks a lot for your input on this! I've added the last changes. I think the discussion has brought a lot of improvements. I think we're good to merge this and subsequently cut a new release - what do you think?

DivineDominion:
merge it! I'll experiment with it in my app right away and see what comes to mind (:

Dimillian:
It's not like I'm using it in a production app for sometimes now :D

PR #205, opened 2017-02-07
Suppress Xcode 8.3 warning

mjarvis:
Compiler warns `Expression implicitly coerced from ‘Any?’ to Any`
We’re safe to force unwrap this (as the variable definition is force unwrapped in the first place)

### DIFF
--- a/ReSwift/CoreTypes/Store.swift
+++ b/ReSwift/CoreTypes/Store.swift
@@ -28,6 +28,11 @@ open class Store<State: StateType>: StoreType {
             subscriptions.forEach {
                 // if a selector is available, subselect the relevant state
                 // otherwise pass the entire state to the subscriber
+                guard let state = state else {
+                    // State should never be nil after `ReSwiftInit`
+                    preconditionFailure("Unexpectedly found `nil` while unwrapping state")
+                }
+
                 $0.subscriber?._newState(state: $0.selector?(state) ?? state)
             }
         }
@@ -112,12 +117,10 @@ open class Store<State: StateType>: StoreType {
         state = newState
     }
 
-    @discardableResult
     open func dispatch(_ action: Action) -> Void {
         dispatchFunction(action)
     }
 
-    @discardableResult
     open func dispatch(_ actionCreatorProvider: @escaping ActionCreator) -> Void {
         if let action = actionCreatorProvider(state, self) {
             dispatch(action)


Ben-G:
Thanks for this! Would you be OK seeing if we come to a conclusion on #203 as that PR addresses this warning as a side-effect of restructuring the subscription?

mjarvis:
@Ben-G I think it would be beneficial to merge this in the meantime, as this also addresses some new warnings regarding `@discardableResult` on a `Void` return.

ReSwift/CoreTypes/Store.swift
```diff
@@ -28,7 +28,7 @@ open class Store<State: StateType>: StoreType {
             subscriptions.forEach {
                 // if a selector is available, subselect the relevant state
                 // otherwise pass the entire state to the subscriber
-                $0.subscriber?._newState(state: $0.selector?(state) ?? state)
+                $0.subscriber?._newState(state: $0.selector?(state) ?? state!)
```
tonyarnold: I'm only new to the codebase, so please excuse me if you already have defined standards around this, but it might be worthwhile using a `guard` with a `preconditionFailure()` here instead of the forced unwrap (`!`), so that you can provide a meaningful error message if the user of this API ever hits a condition where `state` is nil.

Something like:

```swift
guard let state = $0.selector?(state) ?? state else {
    preconditionFailure("Unexpectedly found `nil` while unwrapping state") // This could obviously be more descriptive
}

$0.subscriber?._newState(state: state)
```

mjarvis: @tonyarnold actually it seems my change here is wrong. Your comment prompted me to look at little farther into the behaviour of the app and it appears it should be doing `?? state as Any`

I also think theres a larger issue of https://github.com/ReSwift/ReSwift/pull/177 changing reducer to allow returning nil, but that not getting propagated to subscribers.

I'll update my PR

mjarvis: actually thats not true either... Looking into it further. I've got a test case with a nil reducer that passes with the current code `$0.subscriber?._newState(state: $0.selector?(state) ?? state)`, but crashes with both `!` and `as any`...


```
    /**
     it allows the app reducer to return nil
     */
    func testAllowsReducerReturningNil() {
        let reducer: Reducer<TestAppState> = { action, state in
            return nil
        }
        store = Store(reducer: reducer, state: TestAppState())
        let action = SetValueAction(3)
        let subscriber = TestSubscriber()

        store.subscribe(subscriber)
        store.dispatch(action)

        XCTAssertNil(store.state)
        XCTAssertEqual(subscriber.receivedStates.count, 1)
        guard let last = subscriber.receivedStates.last else {
            XCTFail("Received state count 1, but could not fetch last")
            return
        }
        XCTAssertNil(last.testValue)
    }
```

mjarvis: @Ben-G I think theres a larger issue at play here introduced by #177 . Was it intentional to allow reducers to return nil? If yes we need to adjust `var state: State!` -> `var state: State?` and make other changes appropriately. If not we should switch reducer back to returning non-nil (which matches the old behaviour) and lets it be safe to force-unwrap the state since it can only be nil before `dispatch(ReSwiftInit())`

Ben-G: @tonyarnold @mjarvis Uh good catch! My memory is failing me and I can't find a reference as to why the reducer return type was changed to be optional. I'm pretty sure it wasn't intentional.

The good news is that we've been slow to ship a new release, so we could undo the change without breaking backwards compatibility.

I can't think of a case where it would be useful to have the entire app state be optional after `ReSwiftInit`.

mjarvis: https://github.com/ReSwift/ReSwift/pull/211

mjarvis:
Rebased & updated unwrap based on feedback from @tonyarnold at https://github.com/ReSwift/ReSwift/pull/205#discussion_r101688570

dagio:
Let's merge it and release it as `v3.0.1` ?

PR #206, opened 2017-02-08
remove single param subscribe func 

ekurutepe:
Remove single parameter `subscribe` function and replace it with a default `nil` selector parameter.

### DIFF
--- a/ReSwift/CoreTypes/StoreType.swift
+++ b/ReSwift/CoreTypes/StoreType.swift
@@ -37,15 +37,6 @@ public protocol StoreType: DispatchingStoreType {
      state in this store changes.
 
      - parameter subscriber: Subscriber that will receive store updates
-     */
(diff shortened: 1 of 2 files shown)


ReSwift/CoreTypes/StoreType.swift
```diff
@@ -37,15 +37,6 @@ public protocol StoreType: DispatchingStoreType {
      state in this store changes.
 
      - parameter subscriber: Subscriber that will receive store updates
-     */
-    func subscribe<S: StoreSubscriber>(_ subscriber: S) where S.StoreSubscriberStateType == State
```
mjarvis: I would be inclined to leave this in place since we can't put default parameter values in a protocol. Otherwise areas that use `StoreType` as a generic can't use `subscribe` without manually specifying a `nil` selector.

ekurutepe: right, makes sense. didn't think of that. closing PR.

PR #211, opened 2017-02-20
Revert nullable return for Reducer from 907e2f6

mjarvis:
The Store's state should never be nil after `ReSwiftInit`. Allowing a reducer to return nil contradicts that, and goes against the interface of the store which declares `var state: State!`.

Since this code was never released we can safely revert this change without breaking backwards compatibility.

Based on the conversation from https://github.com/ReSwift/ReSwift/pull/205#discussion_r101688570


mjarvis:
👍 Will merge once CI passes

PR #213, opened 2017-03-06
Resolve all SwiftLint warnings

gin0606:


Ben-G:
Yay! Thanks for these fixes @gin0606. We're inviting all contributors to become members of the ReSwift org, that makes future contributions easier if you're interested. If you're not interested, feel free to ignore the invite!

PR #217, opened 2017-03-13
Replace optional DispatchFunction Middleware by plain DispatchFunction

dimazen:
**Motivation:**
It turns out that there is no need to have an optional dispatch function for the `Middleware`. The only case where it might cause issues (crashes) is when the `Store` gets deallocated and the `Middleware` tries to dispatch a new action. For example this might be a result of the async networking operation or so. However in the idiomatic redux we do not have a temporary store rather a permanent one. Therefore it is very unlikely that store will be deallocated without a middlewares. 

**Impact on the existing code:**
Users will have to replace `dispatch?(action)` by `dispatch(action)` in their middlewares. Xcode detects such things easily. 

**Alternatives considered:**
We can use `weak` instead of `unowned` in the Store during construction of the dispatch function. In addition to this we may also want to print a message during attempt to talk to the deallocated store. 
But going back to the motivation I've decided to go with unowned, because of very low possibility for such case.

### DIFF
--- a/ReSwift/CoreTypes/Middleware.swift
+++ b/ReSwift/CoreTypes/Middleware.swift
@@ -11,4 +11,4 @@ import Foundation
 public typealias DispatchFunction = (Action) -> Void
 public typealias GetState = () -> StateType?
 public typealias Middleware =
-    (DispatchFunction?, @escaping GetState) -> (@escaping DispatchFunction) -> DispatchFunction
+    (@escaping DispatchFunction, @escaping GetState) -> (@escaping DispatchFunction) -> DispatchFunction
--- a/ReSwift/CoreTypes/Store.swift
+++ b/ReSwift/CoreTypes/Store.swift
@@ -50,11 +50,10 @@ open class Store<State: StateType>: StoreType {
         // Wrap the dispatch function with all middlewares
         self.dispatchFunction = middleware
             .reversed()
-            .reduce({ [unowned self] action in
-                return self._defaultDispatch(action: action)
-            }) { [weak self] dispatchFunction, middleware in
-                let getState = { self?.state }
-                return middleware(self?.dispatch, getState)(dispatchFunction)
+            .reduce({ [unowned self] action in return self._defaultDispatch(action: action)
+            }) { [unowned self] dispatchFunction, middleware in
+                let getState: GetState = { self.state }
+                return middleware(self.dispatch, getState)(dispatchFunction)
         }
 
         if let state = state {
--- a/ReSwiftTests/StoreMiddlewareTests.swift
+++ b/ReSwiftTests/StoreMiddlewareTests.swift
@@ -42,7 +42,7 @@ let dispatchingMiddleware: Middleware = { dispatch, getState in
         return { action in
 
             if var action = action as? SetValueAction {
-                dispatch?(SetValueStringAction("\(action.value)"))
+                dispatch(SetValueStringAction("\(action.value)"))
             }
 
             return next(action)
@@ -60,7 +60,7 @@ let stateAccessingMiddleware: Middleware = { dispatch, getState in
             // avoid endless recursion by checking if we've dispatched exactly this action
             if appState?.testValue == "OK" && stringAction?.value != "Not OK" {
                 // dispatch a new action
-                dispatch?(SetValueStringAction("Not OK"))
+                dispatch(SetValueStringAction("Not OK"))
 
                 // and swallow the current one
                 return next(StandardAction(type: "No-Op-Action"))
@@ -111,7 +111,7 @@ class StoreMiddlewareTests: XCTestCase {
     }
 
     /**
-     it middleware can access the store's state
+     it can access the store's state
      */
     func testMiddlewareCanAccessState() {
         let reducer = TestValueStringReducer()


ReSwift/CoreTypes/Middleware.swift
```diff
@@ -11,4 +11,4 @@ import Foundation
 public typealias DispatchFunction = (Action) -> Void
 public typealias GetState = () -> StateType?
 public typealias Middleware =
-    (DispatchFunction?, @escaping GetState) -> (@escaping DispatchFunction) -> DispatchFunction
+    (@escaping DispatchFunction, @escaping GetState) -> (@escaping DispatchFunction) -> DispatchFunction
```
dimazen: To the reviewers: I'm not sure how to format it best to not exceed 100 symbols limit: 

```
public typealias Middleware =  (
    @escaping DispatchFunction, @escaping GetState
) -> (@escaping DispatchFunction) -> DispatchFunction
```

Any suggestions?

Ben-G: I would recommend:

```
public typealias Middleware =  (
    @escaping DispatchFunction, 
    @escaping GetState
) -> (@escaping DispatchFunction) -> DispatchFunction
```

Ben-G:
@dimazen I definitely like idea of removing the optional types from the middleware API! I also agree that `unowned` shouldn't cause a problem when the ReSwift store is used like in an idiomatic Redux pattern.

However, I wouldn't want to limit users of the library to use the store differently. That might cause the edge case you described above. I would prefer to handle that gracefully, as I think we will see someone running into this issue in the future. I think the right behavior of the middleware when the store is no longer present is simply to ignore and discard the dispatched action.

We could do that somewhat like this:

```
self.dispatchFunction = middleware
            .reversed()
            .reduce({ [unowned self] action in
                return self._defaultDispatch(action: action)
            }) { [weak self] dispatchFunction, middleware in
                guard let `self` = self else {
                    return { _ in }
                }

                let getState = { self.state }
                return middleware(self.dispatch, getState)(dispatchFunction)
        }
```

What are your thoughts on that?

mjarvis:
@Ben-G your example there has a problem where an async middleware would retain the store until it completes (Since it retains `dispatch` which retains `self`)

If you want dispatch to "do nothing" but allow the store to be released, there are two options:
1. Maintain the `[weak self]` and optional dispatch.
2. Provide a separate closure for the dispatch function (example below)

```swift
self.dispatchFunction = middleware
            .reversed()
            .reduce({ [unowned self] action in
                return self._defaultDispatch(action: action)
            }) { dispatchFunction, middleware in // since reduce is sync we don't need weak here
                let dispatch = { [weak self] in self?.dispatch($0) }
                let getState = { [weak self] in self?.state }
                return middleware(dispatch, getState)(dispatchFunction)
        }
```

dimazen:
Hey @Ben-G! 

I agree that `unowned` adds potential crash. Originally I was designing code similar to this:

```swift
let defaultDispatch: DispatchFunction = { [weak self] action in
    if self == nil { ReSwiftDeallocatedStoreDispatch(action: action) }
    self?._defaultDispatch(action: action)
}
let middlewareDispatch: DispatchFunction = { [weak self] action in
    if self == nil { ReSwiftDeallocatedStoreDispatch(action: action) }
    self?.dispatch(action)
}
let getState: GetState = { [weak self] in
    if self == nil { ReSwiftDeallocatedStoreGetState() }
    return self?.state
}
self.dispatchFunction = middleware
    .reversed()
    .reduce(defaultDispatch) { dispatchFunction, middleware in
        return middleware(middlewareDispatch, getState)(dispatchFunction)
}
```

Where contents of mentioned `ReSwiftDeallocatedStoreDispatch` is: 
```
private func ReSwiftDeallocatedStoreDispatch(action: Action) {
    print(
        "Attempt to dispatch an action <\(action)> to the deallocated store.\n" +
        "Make a symbolic breakpoint at \(#function) to catch this in the debugger.\n"
    )
}
```

It makes code a little bit more complicated but more developer-friendly :) 
What do you think?

dimazen:
This function `ReSwiftDeallocatedStoreDispatch` gets invoked only for the deallocated store. Developer can put a symbolic breakpoint and catch the place where he talks to the dead store.

dimazen:
@Ben-G what do you think? :)

Ben-G:
@dimazen I actually really like @mjarvis suggested approach as it keeps the store simple. I don't think we need to specifically handle the case in which a store gets deallocated and an action gets dropped.

I've tried out his suggestion on a branch, you can see the result here: https://github.com/ReSwift/ReSwift/compare/master...benji/stop-optional-dispatch

If we all agree than we should go forward with that solution. But feel free to just pull it into this PR; after all it was your initiative :)

dimazen:
@Ben-G Looks good to me. I don't see any reason to intercept your branch :) Probably we can close this PR and merge your branch directly :)

Ben-G:
@dimazen That works for me, too :) Still consider this a great first contribution! Like all contributors I've invited you to become a member of the ReSwift org. This will make it easier for you to contribute in future in case you are interested!

Ben-G:
Closing in favor of #225.

dimazen:
In case there will be more significant contributions from my side - I accept your invitation without any hesitation :)

Issue #207, opened 2017-02-12
Add SwiftLint to CI script


mokagio:
At this point in time ReSwift is failing to build on machines using SwiftLint version 0.16.1. #202 addresses the issue, but I think we can do better than that.

In [a comment](https://github.com/ReSwift/ReSwift/pull/202#issuecomment-277282084) on that same PR @DivineDominion says:

> I don't have a project on CI to check what would change.

That made me think... would it be good to have our CI script fetching the latest SwiftLint, so that we can ensure that the project would build under it? 🤔 

I can see that we already have [Hound](https://houndci.com/) integrated in the PRs. That is awesome! The two things can totally live together.

- Hound 👉 make inline comments in the PRs for style violations
- SwiftLint in CI 👉 ensure the framework build process is compatible with the latest version of SwiftLint

We could even setup a [Travis CI cron job](https://docs.travis-ci.com/user/cron-jobs/) to run a build on `master` daily, so that we'd know about possible incompatibilities with new SwiftLint versions within 24 hours from their release.

mokagio:
If the idea makes sense I'd be happy to start working on a PR for it. I've done it before and shouldn't involve too much work 👍 .

DivineDominion:
👍

If you could take a few notes about the stuff you do, maybe even I can follow along and help taking care of this stuff in the future 📚

mjarvis:
> We could even setup a Travis CI cron job to run a build on master daily, so that we'd know about possible incompatibilities with new SwiftLint versions within 24 hours from their release

I think I'd be more inclined to determine a solution for pinning the version of SwiftLint that is used here, so that we don't have issues like this. (IE, what if a user checks out an old commit? In my opinion it should still be able to compile without modification)

mokagio:
@mjarvis that would be an ideal solution, although I don't think it's possible with how SwiftLint works at the moment, at least for my understanding of it.

Have you got any suggestion?

The only way that I see to ensure that changes to SwiftLint won't interfere with how ReSwift builds would be to "vendor" it in the repo. With that I mean fetch SwiftLint as a submodule, and use that version rather than the one that should be in the user's path in our build process.

Submodules come with their own bag of troubles, so I would not recommend this option 😞 .

mjarvis:
@mokagio I think the best way to pin swiftlint would be to use cocoapods to bring it in, which would lock the version in `Podfile.lock`, and allow it to be manually updated via `pod update`.

[...]

PR #218, opened 2017-03-16
Pin SwiftLint using CocoaPods

mokagio:
This PR addresses [the action item](https://github.com/ReSwift/ReSwift/issues/207#issuecomment-281547836) resulting from the conversation in #207.

It fetches SwiftLint via CocoaPods, so that we can control the version used by all developers and refs, and more importantly it only runs SwiftLint if it has installed in the project.

When dealing with CocoaPods, or any other Ruby tool, I would normally set it up using a Bundler. I'm not sure what we want to do in this instance though. Would it be overkill? (_I don't think so_).

Also note that because of the way `pod install` works I had to wrap the SwiftLint via CocoaPods setup in a script. There is no way to run `pod install` and only fetch libraries, so what I'm doing in this script is running it to fetch the SwiftLint binary, and then undoing all the changes to the project file made by it.

If anyone can come up with a cleaner solution for the integration I'd love to talk about it. I'm not very happy with it, it is not robust, and might not be straightforward for new developers.


.gitignore
```diff
@@ -35,13 +35,11 @@ xcuserdata
...
-# We recommend against adding the Pods directory to your .gitignore. However
-# you should judge for yourself, the pros and cons are mentioned at:
-# https://guides.cocoapods.org/using/using-cocoapods.html#should-i-check-the-pods-directory-into-source-control
-#
-# Pods/
+# While ReSwift has no dependencies we are still using CocoaPods, but only
+# to ensure developers use the same binary version of SwiftLint. Since
+# this is just a development dependency it is ok to ignore it, and rely on
+# developers to generate it when required.
+Pods/
```
mokagio: I think it makes sense to ignore the Pods folder for the sake of keeping the repo lean.

BuildPhases/run_swiftlint.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+set -e
+
+swiftlint=$SRCROOT/Pods/SwiftLint/swiftlint
+
+if [ -f $swiftlint ]; then
+  $swiftlint
+else
+  echo 'warning: Cannot find SwiftLint in the expected location. If you are developing ReSwift please install it via using `./Scripts/install_swiftlint.sh`. If you are building ReSwift as a dependency of your project feel free to ignore this.'
```
mokagio: I wanted to make this a multiline string for readability, but Xcode only prints the first line of warnings and errors, afaik.

Podfile
```diff
@@ -0,0 +1,7 @@
+# CocoaPods is used only to fetch and pin the version of SwiftLint that we
+# want to use in the project, we are defining it for ReSwift-iOS only
+# because the Podfile DSL (at the time of CP 1.2.0) requires pods to be
+# defined within a target.
```
mokagio: This is the error:

```
[!] The dependency `SwiftLint (= 0.16.1)` is not used in any concrete target.
```

Scripts/install_swiftlint.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+set -e
+
+# Run pod install to fetch the SwiftLint binary
+pod install
+
+# Undo all the changes made by CocoaPods
+git checkout ReSwift.xcodeproj/project.pbxproj
```
mokagio: This could backfire, discarding other changes to this file before running the script and not committed.

I did play around with stashing before running `pod install`, but that can backfire as well, because if you don't have any changes and an empty stash, when you'll try to `pop` you'll get a "No stash found" error.

🤔

mjarvis:
What happens if you specify swiftlint in the tests target for cocoapods, and allow it to modify the project file? The project file is only used by people developing on ReSwift, so its not such a big deal to have it be modified during the initial pod install.

mjarvis:
I'm still not certain I understand whats so bad about including swiftlint in a real target committing the xcworkspace & project changes. I feel like thats much more simple and easy to understand for a developer coming in, say, if they want to add another development dependency (some mocking framework or something), instead of having revert a bunch of auto-discarded or ignored things from this.

mokagio:
Thanks for the feedback @mjarvis 😄 .

In f59723b I added a new dummy target where to integrate SwiftLint. I think a dedicated target is better than one of the current ones because this way all the bloat goes into it rather than in an actual target.

While I don't like the project being a bit dirty because of this I do think that this is a more robust solution. Thanks for suggesting it 😃 👍 .

mokagio:
> I'm still not certain I understand whats so bad about including swiftlint in a real target committing the xcworkspace & project changes.

From my point of view, it adds unnecessary build phases and complexity to a target when it is not necessary.

> I feel like thats much more simple and easy to understand for a developer coming in, say, if they want to add another development dependency (some mocking framework or something), instead of having revert a bunch of auto-discarded or ignored things from this.

I think this assumes that ReSwift should use CocoaPods to manage it's dependencies. Is this actually the case? [ReSwift-Router](https://github.com/ReSwift/ReSwift-Router) uses Carthage for example.

Using a dedicated target for SwiftLint seems more self-contained to me. Also how would you choose which target to put it in then? `ReSwift-iOSTests`, `ReSwift-macOSTests`, `ReSwift-tvOSTests`? Why using one rather than the other?

> instead of having revert a bunch of auto-discarded or ignored things from this.

I think you have a good point on the having to revert stuff though. I'll remove the `xcworkspace` from the `.gitignore`. A future developer might actually need the workspace (can't see why atm but one never knows 😄 ) and waste time not understanding why their changes are not appearing under Git.

Ben-G:
> I think this assumes that ReSwift should use CocoaPods to manage it's dependencies. Is this actually the case? ReSwift-Router uses Carthage for example.

This brings up a good overall point. Right now Carthage is the recommended way of working on ReSwift. While I'm not opposed to changing that I think it's a slightly larger discussion, that should probably include @DivineDominion and maybe @agentk?

DivineDominion:
As long as nothing gets broken, I don't really care if the folder structure changes to accommodate CocoaPods :)

Ben-G:
@mokagio Tried out this branch and think it's a great improvement to have a pinned `SwiftLint` version! Thanks for putting in extra effort to avoid bloat of the framework target.

I would suggest, as a last step, to add some info to the `README.md` to the *Checking out Source Code* section, that mentions to run `pod install` to get latest version of SwiftLint, but otherwise just develop with `.xcodeproj`. Once that is done, this is good to merge!

mokagio:
@mjarvis good idea removing the version specifier from the `Podfile`, one needs to _intentionally_ run `pod update`, so the pinning on that file was much of a muchness 👍 .

Do you mind if I take care of adding the info suggested by @Ben-G and rebase this branch on `master` to remove the merge commit?

mjarvis:
@mokagio Feel free to take care of adding the info. I'd suggest not rebasing, though, and instead we can take advantage of github "Squash and merge" which will give us a single commit representing this changeset. (And will leave the history intact in the PR)

mokagio:
Thanks @mjarvis for the feedback and review along this unfortunately long lived PR.

I'm good to "Squash and merge" right?

mjarvis:
Yes I think Squash and merge is the best option given the long-winded history. I don't think that we require that in the actual git history, and its maintained here in the PR if needed.

mokagio:
💥 done.

I have feelings about "Squash and merge", but yeah this change is quite atomic and it can fit nicely into a single commit 👍 . What's actually valuable is the discussions we had between the #207 and this PR, which wouldn't have been in the Git history anyway.

Cheers 🍻

PR #221, opened 2017-03-20
Update README example to newer reducer format

mjarvis:
In #177, `Reducer` was changed from a protocol, requiring the user to create a struct to implement it, to a typealias, allowing one to simply write a function conforming to it. This updates the Readme example to use the new typealias based reducer instead of a struct conforming to the `Reducer` protocol.

### DIFF
--- a/Docs/Getting Started Guide.md
+++ b/Docs/Getting Started Guide.md
@@ -81,22 +81,18 @@ You can do this implementing a top-level reducer that conforms to the `Reducer` 
 Here's an example in which we construct a new state, by calling sub-reducers with different sub-states:
 
 ```swift
-struct AppReducer: Reducer {
-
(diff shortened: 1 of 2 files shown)


mjarvis:
There is also another format for creating a reducer that one could use:
```swift
func counterReducer() -> Reducer<AppState> {
    return { action, state in
        var state = state ?? AppState()

        switch action {
        case _ as CounterActionIncrease:
            state.counter += 1
        case _ as CounterActionDecrease:
            state.counter -= 1
        default:
            break
        }

        return state
    }
}

let mainStore = Store<AppState>(
    reducer: counterReducer(),
    state: nil
)
```

Does anyone have a preference as to which format should be provided as example?

Ben-G:
Thanks a lot for taking care of docs update! This looks good to me!

PR #226, opened 2017-03-23
Make Middleware's `getState` generic

mjarvis:
This allows for us to specify the state type that middleware uses, providing compile-time assurance of state contents, instead of requiring the user to deal with conditional type-casting during runtime.

This seems to have an issue compiling with Xcode 8.2, but works fine in 8.3.

- [x] Update documentation
- [x] Figure out why the build times out in xcode 8.2

### DIFF
--- a/Docs/Getting Started Guide.md
+++ b/Docs/Getting Started Guide.md
@@ -247,7 +247,7 @@ Let's take a look at a quick example that shows how ReSwift supports Redux style
 The simplest example of a middleware, is one that prints all actions to the console. Here's how you can implement it:
 
 ```swift
-let loggingMiddleware: Middleware = { dispatch, getState in
+let loggingMiddleware: Middleware<Any> = { dispatch, getState in
     return { next in
         return { action in
             // perform middleware logic
@@ -259,6 +259,8 @@ let loggingMiddleware: Middleware = { dispatch, getState in
     }
 }
 ```
+The generic in middleware refers to the return type in `getState`, and needs to be compatible with the `State` associated type in your `Store`.
+
 You can define which middleware you would like to use when creating your store:
 
 ```swift
--- a/ReSwift/CoreTypes/Middleware.swift
+++ b/ReSwift/CoreTypes/Middleware.swift
@@ -9,6 +9,6 @@
 import Foundation
 
 public typealias DispatchFunction = (Action) -> Void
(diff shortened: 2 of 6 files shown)


Ben-G:
@mjarvis Curious: have you thought about cases in which one middleware has a generic type argument of `StateType` and another has one of `SomeSpecificStateType`. I think you wouldn't be able to combine these middlewares without resorting to some sort of type-erasure?

mjarvis:
You can combine them as long as the holding type is more concrete:

```
protocol StateType {}
protocol Foo {}

struct AppState: StateType, Foo {}

typealias Middleware<State> = (State) -> Void

func openMiddleware<S>(_ state: S) -> Void {}
func protocolMiddleware<S: Foo>(_ state: S) -> Void {}
func structMiddleware(_ state: AppState) -> Void {}

class Store<State: StateType> {
    let middlewares: [Middleware<State>]

    init(middlewares: [Middleware<State>]) {
        self.middlewares = middlewares
    }
}

let store = Store<AppState>(
    middlewares: [
        openMiddleware,
        protocolMiddleware,
        structMiddleware,
    ]
)

store.middlewares.count // 3
```

mjarvis:
If you prefer the variable assignment style of middleware:

```
let openMiddleware: Middleware<Any> = { _ in }
let protocolMiddleware: Middleware<Foo> = { _ in }
let structMiddleware: Middleware<AppState> = { _ in }
```
works just as well

mjarvis:
As suspected, this failed to build on travis... I suspect this work will have to wait until ReSwift can require Xcode 8.3.

Ben-G:
Ah, yes. This is why you were suggesting functions in the first place; their generic type can remain unspecialized which enables this. Looks like a great idea!

Ben-G:
Ah, yes. This is why you were suggesting functions in the first place; their generic type can remain unspecialized which enables this. Looks like a great idea!

Ben-G:
Also the build deadlock in Xcode 8.2 is really interesting...

mjarvis:
Inlining `GetState` into the `Middleware` typealias appears to fix the indexing & compilation on my computer -- lets see if it resolves it on travis.

Ben-G:
@mjarvis Looks good! Rebase and merge?

Issue #150, opened 2016-09-11
Action vs ActionCreator vs AsyncActionCreator 


ColinEberhardt:
Having come from using redux, I have a few questions regarding the ReSwift types and how they map to redux. 

Within redux, Actions are simply objects with the correct 'shape', i.e. a type and a payload, and ActionCreators are functions that create these Actions.

With that in mind, I do wonder whether the ReSwift Action is actually an ActionCreator? Looking at the code for a simple Action that adds an item to a to-do list:

```
struct AddToDo: Action {
    let text: String

    init(text: String) {
        self.text = text
    }
}
```

This is both an Action (it has a type and a payload), and an ActionCreator (it has an initialiser that constructs the Action).

This might sound very pedantic, but here's where I think ReSwift starts to deviate ...

The ReSwift `ActionCreator` type has state and store parameters. This is quite different from redux:

> In general accessing state in action creators is an anti-pattern and you should avoid this when possible. 

from Dan Abramov - http://stackoverflow.com/questions/35667249/accessing-redux-state-in-an-action-creator

Again, this might all sounds a little pedantic, but it does lead to some real confusion when trying to understand how to dispatch actions asynchronously. I'd expect this to be the role of the `AsyncActionCreator`, however, you can also do this (more easily) via `ActionCreator` where you returns `nil`, and instead dispatch to the store. I've noticed this being done in the following example:

https://github.com/Ben-G/Meet/blob/master/Meet/Meet/Actions/TwitterAPIActionCreator.swift#L58

I do wonder whether ReSwift would also benefit from using the same pattern as redux where, rather than providing functions with the store, they instead provide a 'view' onto the store which just allows them to dispatch or invoke `getState`

Sorry - a bit of a messy brain-dump here. I'd be happy to turn this into something more concrete if you're interested?

Ben-G:
@ColinEberhardt just confirming that I did read this post. Had a crazy week due to Xcode 8 / iOS 10 related fixes. You're bringing up good points. Want to get back to them latest over weekend.

ColinEberhardt:
No worries, thanks for letting me know!

DivineDominion:
Closing this with #231 with the recent addition of ReSwift-Thunk to the ecosystem (e.g. #372 and #378)

PR #231, opened 2017-03-28
Generic ActionCreator

mjarvis:
This splits ActionCreator out of `StoreType` and makes it a stand-alone
generic typealias. This allows one to define an action creator without the
context of the store its being called on, and lets you use that action
creator on any conforming store. This allows more isolated action creators
that need only know about some sub-state.

See `testAcceptsConformingActionCreators` for an example.

A requirement for this is also removing the `Store` argument from `ActionCreator`.

I can't see a valid use case for providing the store to an action creator, anyways.
Including it seems to encourage bad practices such as accessing the state
via the store, or dispatching additional actions from an action creator,
which I feel should be discouraged.

### DIFF
--- a/Docs/Getting Started Guide.md
+++ b/Docs/Getting Started Guide.md
@@ -224,7 +224,7 @@ Just like an `Action` a `ActionCreator` function can be dispatched to the store.
 
 An `ActionCreator` has the following type signature:
 ```swift
-typealias ActionCreator = (state: State) -> Action?
+typealias ActionCreator<S: StateType> = (state: S) -> Action?
 ```
 
 A very simple example of an `ActionCreator` might be:
--- a/Docs/Getting Started Guide.md
+++ b/Docs/Getting Started Guide.md
@@ -220,16 +220,16 @@ An important aspect of adopting `ReSwift` is an improved separation of concerns.
 
 The triggering of actions should always be as simple as possible, we want to avoid any sort of complicated business logic in the view. However, in some cases it can be complicated to decide whether an action should be dispatched or not. Instead of checking the necessary state directly in the view or view controller, you can use `ActionCreator`s to perform a conditional dispatch.
 
-Just like an `Action` a `ActionCreator` function can be dispatched to the store. An `ActionCreator` takes the current application state, and a reference to a store and might or might not return an `Action`.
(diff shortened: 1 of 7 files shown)


danielmartinprieto:
Hi! Regarding this topic, have you seen this issue created some time ago? https://github.com/ReSwift/ReSwift/issues/150 I think he makes some very good points that could help shaping the new actions helpers.

mjarvis:
@danielmartinprieto Thanks for the link. I took a look through, at it seems like this PR helps with the issues described there, by not allowing the user to subvert the intention of an actioncreator by dispatching in it.

> In general accessing state in action creators is an anti-pattern and you should avoid this when possible.

I think the idea of an action creator is different in redux than in reswift. In redux an action creator is literally the initializer for an action (as actions aren't strongly typed), but in reswift we have strongly-typed actions via structs, which provide their creator inline. Thus, the action creator described here is actually a separate helper, which allows the user to choose whether to dispatch an action based on some given state.

I feel like there is some underlying information in that issue that could provide a better way to do that (via a simple action, reducer, then something that does the work separately based on the state?) but I'm having a hard time visualizing it without an example.

My basic use case for the ReSwift action creator is an `authorize(request: URLRequest)` type function, which adds an authorization header if the state has an authorization object, or dispatches a log-able error otherwise.

mjarvis:
@rpassis Yes I believe the same should be done for AsyncActionCreator -- however I don't ever use that in my projects, so I didn't want to adjust it without making sure changing it won't break desired use cases.

DivineDominion:
@mjarvis to continue the discussion here, see: https://github.com/ReSwift/ReSwift/issues/303#issuecomment-343778977

@danielmartinprieto's point is:

> Regarding the new AsyncActionCreator type, its API is really close to the middleware's one: you have access to the current state and the dispatch function, but then its usage is spread across the whole app, instead of having a single place where you perform the side effects using a certain IO entity.

A follow-up proposal, in my own words: extract ActionCreator types out of the core because one might rarely need them in practice and _should_ (!) stick to simple actions and middleware.

DivineDominion:
@mjarvis when the thunk middleware "plugin" is ready, how'd you like us to progress with this?

mjarvis:
I believe `ReSwift-Thunk` will allow us to deprecate & remove `ActionCreator` and `AsyncActionCreator`, in which case this PR is irrelevant.

PR #239, opened 2017-04-27
[Release] Stop Producing Binaries

Ben-G:
Using precompiled binary frameworks is not supported by Apple. Further,
the pre-built binaries are being built with the version of Swift that is
current at the time of the release. This leads to problems of Swift version
mismatches when trying to install ReSwift via Carthage.
I suggest we stop producing and publishing binaries.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,17 +44,3 @@ after_success:
     test $TRAVIS_BRANCH == "master" &&
     test $UPDATE_DOCS == "true" &&
     .scripts/update-gh-pages
-
-before_deploy:
-  - carthage build --no-skip-current
-  - carthage archive $FRAMEWORK_NAME
-
-deploy:
-  provider: releases
-  api_key:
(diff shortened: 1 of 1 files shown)


mjarvis:
> problems of Swift version mismatches when trying to install ReSwift via Carthage

Does Carthage intend for the user to deal with that using `--no-use-binaries`?

Ben-G:
@mjarvis It does; but as per the discussion above I actually don't agree that using binaries is a good default. 

If we want to keep distributing binaries, we would IMO need to create a new release for every new version of Swift immediately + update all other repos in the ReSwift org. Otherwise the build instructions for Carthage on these projects will be incorrect (they lead to build errors). That's an issue I just ran into myself.

Besides requiring some effort to automate that I think it would create a lot of noise around releases.

Does that make sense? Happy to discuss further!

DivineDominion:
Off with their heads! +1 Good idea.

PR #242, opened 2017-05-03
Add Subscription `skip(when:)` and `only(when:)`

mjarvis:
These subscription filtering methods are provided as convenience methods for `skipRepeats`.

1. `skip(when:)` is a direct forward to `skipRepeats` and is simply provided as a more swift-like naming option.
2. `only(when:)` is the inverse to `skipRepeats`, and is provided for specifying when the user wants the subscription to notify, as opposed to when to skip notifying.

Both of these functions are technically "sugar", however I feel like they add value to the library by making the available use cases of `skipRepeats` more obvious to users.

For example, a user could chain calls to these, such as `$0.select({ $0.child }).skip(when: { $0.identifier == $1.identifier }).only(when: { $1.name.lowercased.hasPrefix("a") })` which would notify the subscriber when the child changes, and the new child's `name` starts with "a"

Also adds a bit more documentation to `skipRepeats`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# Upcoming
+
+**Other**:
+- Add Subscription `skip(when:)` and `only(when:)` (#242) - @mjarvis
+
 # 4.0.0
 
 *Work in Progress*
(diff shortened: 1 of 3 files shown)


danielmartinprieto:
Hi! Looks cool.

I have a question. If my state conforms to Equatable and I use `skip(when:)`, it overrides the original skipRepeats or it happens after that?

DivineDominion:
👍 I like `skip(when:)` better than `skipRepeats`. Maybe we could rename this in the future?

mjarvis:
@danielmartinprieto thats a good question, actually -- I wonder if we end up with equitability checks when one makes an equatable substate selection & a custom implementation of `skipRepeats`. I'll look into it.

Ben-G:
@mjarvis Would you mind adding a Changelog entry as well?

PR #247, opened 2017-05-14
[Docs] Remove Swift Version from Jazzy

Ben-G:
The incorrect Swift version was causing Jazzy to fail generating docs. We should leave the setting to always use the latest Swift version implicitly.


Ben-G:
@ReSwift/reswift Anyone mind signing off on this so we can test if it works out when building `master`?

DivineDominion:
I added a Rakefile to automate doc generation so folks like me don't have to look this up.

After running jazzy, I get a list of files that are not part of the repo:

```
	Docs/Classes.html
	Docs/Classes/
	Docs/Protocols.html
	Docs/Protocols/
	Docs/Structs.html
	Docs/Structs/
	Docs/badge.svg
	Docs/css/
	Docs/docsets/
	Docs/img/carat.png
	Docs/img/dash.png
	Docs/img/gh.png
	Docs/index.html
	Docs/js/
	Docs/search.json
	Docs/undocumented.json
```

Should we add them to `.gitignore` since the CI tool will probably build that locally?

Ben-G:
@DivineDominion Actually the doc generation is already automated as part of the `.travis.yml` file: https://github.com/ReSwift/ReSwift/blob/master/.travis.yml#L43-L46

The deployment was just breaking due to the incorrect Swift version specified in the jazzy config file. So I think we should look into getting feedback on failed automatic generation of the docs. I don't think we need the `Rakefile` in this case. Does that make sense? 😃

Ben-G:
Oh no, @DivineDominion, I didn't notice that the `Rakefile` was commited on top of this PR. Will remove it with a separate PR. In general I think it would be best to open separate PRs instead of working off of existing ones 😄

DivineDominion:
Oh, sorry for the trouble. Figured this belonged to "fix docs generation" to test docs locally without much hassle, but it did sure feel a bit intrusive.

Ben-G:
No big deal :) I also could have taken another look at the PR before merging 😉 

If you think the Rakefile is still useful for testing locally, we should just open a separate PR :)

DivineDominion:
Right, done! #250

Issue #248, opened 2017-05-15
add release checklist to repository


DivineDominion:
Collecting things that come to mind (and which can be scripted in part):

- write release notes 
    - collect contributors per merged feature
- try to generate docs (how exactly?)

---

Feel free to edit this directly so we have a single point of reference.

Afterwards: 

- [x] add checklist to README or separate file
- [ ] write script to automate building the docs locally and pull list of PRs since last release

DivineDominion:
checklist see #312

mjarvis:
Don't the docs get automatically generated already? There is a CI script that gets run on master pulls: `.scripts/update-gh-pages`

DivineDominion:
I don't know what I was thinking about in 2017 anymore. Maybe this was about easing local doc generation for proof-reading since #250 is referenced. I'm closing this anyway because 18 months without reaction indicates that this isn't super important :)

PR #250, opened 2017-05-16
add Rakefile to automate doc generation

DivineDominion:
Could be a first step towards #248 -- making testing jazzy documentation generation locally easy.


DivineDominion:
@mjarvis @Ben-G do you think an automation script like this is still relevant? Maybe a bot or CI step could do the trick, but I'm not well-versed in these.

mjarvis:
Sorry, I'm not really sure what this does. Is this for generating the docs site? Is this something we should be doing in each PR?

PR #255, opened 2017-05-30
Update README styling

getaaron:
Small change, just inserts a hard return so it doesn't format as one line in HTML. ↩️ 👍


mjarvis:
Thanks @getaaron! Like all contributors I'll invite you to the ReSwift org. That will make future contributions easier in case you are interested! 🎉

PR #256, opened 2017-06-05
Update project for Swift 4.0 / Xcode 9.0

mjarvis:
This is a working branch / PR for Swift 4.0 / Xcode 9.0 migration.
This currently compiles & passes all tests, on all targets, with no warnings or compiler errors (except for a warning regarding the pods project, which we need to wait for Cocoapods to remove).


Ben-G:
Awesome work! Thanks @mjarvis!

mjarvis:
This branch is ready for review! This should be fully backwards compatible for Xcode 8 / swift 3.0 as  there are no changes to the main target.

DivineDominion:
Will have a look in a couple hours!

As @Ben-G would say: do you want to add a CHANGELOG entry? :)

mjarvis:
I don't feel like adding a changelog entry here is necessary, since this contains no changes to ReSwift itself, only to the test target & the development xcode project

mjarvis:
@DivineDominion  Re: Swiftlint, linting runs as a build script during compilation, from a cocoapods integration. The warnings you discovered there only show up when using an updated version of swiftlint. I have a branch to update our integrated swiftlint version. My recommendation is we remove the linting conformance commit from here, and make a separate PR updating the pod swiftlint version & conforming to the new linting rules.

DivineDominion:
I'm fine with that decision 👍 I still think "Swift 4/Xcode 9" conformance should at least be part of the release notes if not the changelog so people know for sure.

mjarvis:
@DivineDominion But the current release already supports Swift 4 and Xcode 9 with no warnings.

DivineDominion:
Yeah sure, my point is just: do users know? Could they hesitate to upgrade? -- Maybe updating the README is enough to be clear anyway

PR #259, opened 2017-06-14
Add test example for double equatable check

mjarvis:
Having a default implementation for `skipRepeats` when the selected state is equatable causes what could be a performance issue when the user implements their own `skipRepeats`.
The original default implementation is still called, even though a default exists.
This could also cause unwanted behaviour if the custom skipRepeats wants to compare something not checked in the default equatable.

### DIFF
--- a/ReSwiftTests/StoreSubscriberTests.swift
+++ b/ReSwiftTests/StoreSubscriberTests.swift
@@ -157,6 +157,35 @@ class StoreSubscriberTests: XCTestCase {
         XCTAssertEqual(subscriber.newStateCallCount, 1)
     }
 
+    /**
+     It avoids default skipRepeats when custom skipRepeats is implemented.
+     */
+    func testCallsEquatableOnceForStateChange() {
+        let reducer = testCustomEquatableReducer
+        let state = TestCustomEquatableAppState()
+        let store = Store(reducer: reducer, state: state)
+        let subscriber = TestFilteredSubscriber<CustomEquatable>()
+        customEquatableCount = 0
+        var customSkipCount = 0
+
+        store.subscribe(subscriber) {
+            $0.select { $0.testValue }.skipRepeats {
+                customSkipCount += 1
+                return $0.value == $1.value
+            }
+        }
+
+        XCTAssertEqual(customEquatableCount, 0)
+        XCTAssertEqual(subscriber.receivedValue.value, 0)
+        XCTAssertEqual(customSkipCount, 0)
+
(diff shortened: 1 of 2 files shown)


mjarvis:
This intentionally fails to demonstrate what may be undesirable behaviour.
My first theory for avoiding this would be to have `select` when Equatable return a different subscription type, which, if the user calls `skipRepeats` on, discards its prior, default skipRepeats implementation.
However, before I go ahead and try and find a solution, I wanted to open the discussion in case others think this behaviour is fine / desirable.

DivineDominion:
I swear we had a user report confusion about this somewhere, recently. Cannot find the issue, though.

It does not make sense to `skipRepeats` eagerly; I'd definitely postpone this to the last possible moment, just like you said:

> My first theory for avoiding this would be to have select when Equatable return a different subscription type, which, if the user calls skipRepeats on, discards its prior, default skipRepeats implementation.

I think something among the lines of a `LazySkipRepeatsSubscription<T>` (Or `LazyFilteredSubscription`) could do the trick. It can use a chain of subscriptions under the hood, delegating `select` calls to itself to the underlying subscriptions _until the selected substate is not skippable anymore._ 

Then more expressibe types can aid; the general `Subscription` becomes a `SubstateSubscription` and `SkipRepeatsSubscription` (or `FilteredSubscription`) etc. so the type system takes care of conveying if something was skipped anywhere, yet.

Writing it down like this makes me realize we are operating on types called "subscription", which conveys the connection to the store is already established but subsequently modified. What about renaming the stuff we call `skipRepeats` and `select` on something more skin to "stream of state updates"? Then a `Subscription` type is just the end result and we can insert as many helper types as we need without causing confusion for subscription setup. `StateUpdate`, `FilteredUpdate<T: StateUpdate>`, `SubstateSelection<T: StateUpdate>`. I'd like to help cook something like this up if it resonates with you.

DivineDominion:
@mjarvis Since we never managed to get #307 to work, do you have an opinion on the renaming of `subscribe` return type names? I can just go ahead and implement it as you proposed to make it quick.

mjarvis:
@DivineDominion I gave this a try a while back but was unable to resolve the problem. You're welcome to take a crack at it.

AlexanderBollbach:
so i ran into this issue and incurred performance hits in my app.  i called `skipRepeats` to insert a specific equality checking closure that was optimized for the specific StoreSubscriber (as one does), but it was never called.   that closure was never invoked.

Issue #258, opened 2017-06-14
[Question] Threading ? 


Dimillian:
I'm noticing an increasing "lag" when a new state is dispatched, because I have a more and more complex state implementing Equatable and using skipRepeat. Is there a way to make my state comparison on a background queue? Would making a custom Subscription a good place?

Dimillian:
Okay silly me. I can just subscribe in a custom queue and my subscription will be called in that queue. Perfect!

DivineDominion:
This is an interesting bottleneck nevertheless. `skipRepeats` wouldn't work well out of the box for larger collections. That's why I compare in my subscriber using not `==` but a "similar enough" operator instead whenever applicable, kind of like a shallow comparison. But you'll have to write that for every type that can have large objects.

For another part of the same app, I added versioning (an incrementing number) to user-initiated requests to discard outdated responses. I wonder if versioning would be a useful for ReSwift states in general through an extension maybe.

Dimillian:
I'll add some more do the discussion after my research: 

So about Equatable, I implement it everywhere it make sense, and for collections I have some very specific implementation case by case, sometimes it make sense to do a deep comparaison on every objects, sometime I need it to be fast, so just comparing ids/string is ok. The downside is that my states and objects are very bloated and it's hard to see what equatable are really doing. So I use the == of Equatable because it's convenient, but it's some case it's more about being similar enough. 

I also have custom operator for merging operations in my reducer, but that's another story. 

About threading, I'm doing everything on a serial high priority background queue now, in short, it looks like that:

```
let FLUX_QUEUE = DispatchQueue(label: "xxx.appName.xxx.store.queue")

extension DispatchQueue {
    public class func flux() -> DispatchQueue {
        return FLUX_QUEUE
    }
}

```

and

```
extension Store {

    open func dispatchQueue(_ action: Action) {
        DispatchQueue.flux().async {
            self.dispatchFunction(action)
        }
    }
}

```

So I use dispatchQueue whenever I would have called a store.dispatch. 

The only downside is that I don't see how to shortcut StoreSubscriber to be called on the main thread, I guess I have so shortcut subscriber first and register my custom class etc... So for now I need to remember to use the main queue in all my new state function call. 

Do you think of a better way of doing that @DivineDominion ?

Dimillian:
@DivineDominion about versionning, you have exactly the same problem as me for my live search! 
[...]

PR #262, opened 2017-06-20
Subscription.select should not surprisingly skipRepeats

DivineDominion:
This basically removes the option to select a substate which is Equatable without skipping repetitions.

> Why would anybody not want to skip repetitions?

I certainly do, but I don't want to do this using the `==` operator, especially not for the huge collections of notes I manage in this one app. Right now I cannot pass a substate along directly. Instead, I have to create a subscription wrapper that selects the substate and passes it on to the real subscriber.

I think this side-effect of `select` hurts, and writing `.select(...).skipRepeats()` shows at the call site what'll be going on. No surprises.


DivineDominion:
So up until this point, my PR takes away the default transformations (which are unintuitive defaults, I'd argue).

As a next step, what do you think about a convenience trigger on `Store` like `isSkippingRepeats` that reintroduces the skip-repeats-upon-subscription behavior as a configuration? Because I do see that there are apps which will want this "on" by default and where the transformations might get on peoples nerves. An alternative, `subscribeSkippingRepeats(...)` sounds clumsy, so I don't suggest that.

As a side note, shallow equatability or no equatability checks at all for subscribers came up recently in #258

mjarvis:
related: https://github.com/ReSwift/ReSwift/pull/259

DivineDominion:
You're right, that's a problem. Coalescing (is this the right term for this case?) `skipRepeats` calls so only the latest substate selection is used sounds useful. However, there'd still be no way to opt-out of the default behavior, no matter how clever the end result is optimized.

If `skipRepeats` wasn't the default, then #259 stops to be a problem from ReSwift's side, too. The user might call `.skipRepeats().select { ... }.skipRepeats()` -- but then that's her problem.

Intellectually, your suggestion is very attractive @mjarvis and I'd love to make that happen, but for the reason of not being able to opt-out at all, I'd favor changing the default behavior here first.

mjarvis:
I like the idea of making a way to opt-out globally to the "default" behaviour of `select` skipping repeats. I wasn't trying to say that my suggestion would be instead of the work you describe, just that it has some relation in that it uses the same APIs. Perhaps there is some cross-over that can resolve both issues.

DivineDominion:
Oh sure, I didn't mean to imply that  :) 

If there is no default `skipRepeats` that gets in the way, what do you think will be missing? 

Do you want to handle stuff like:

    $0.skip(when: ==)
        .select { $0.substate }
        .skip(when: ==)

... to ditch the first `skip` call, too?

mjarvis:
I'm not sure... My initial thoughts are that it should not ditch explicit, user-defined skips.

DivineDominion:
Then this PR's change will throw the problematic baby (ditching automatic skips) out with the bathwater :) When we merge this in, #259 effectively becomes obsolete.

mjarvis:
I suggest that this be made a configuration option on Store -- to allow people to opt-in or out of the behaviour change -- a convenience trigger like you indicated would be fine, Not sure what we'd name it.`subscriptionsAutomaticallySkipRepeats` maybe? I think it should only be configured on init of the store, to avoid users changing it on the fly, which would cause undesirable behaviour since existing subscriptions would not be changed.

DivineDominion:
I like that idea and pushed commits for this change.

Points to discuss:

- [ ] Does the initializer prescription in `StateType` protocol had any benefit at all? I removed it en passant as I pondered which initializer setup would work best.

DivineDominion:
Sorry for pinging, but can you have a look @mjarvis or @Ben-G so that we can merge this rather sooner than later? :)

ReSwift/CoreTypes/Store.swift
```diff
@@ -40,11 +40,29 @@ open class Store<State: StateType>: StoreType {
...
+    ///   before the root reducer.
+    /// - parameter automaticallySkipsRepeats: If `true`, the store will attempt 
+    ///   to skip idempotent state updates when a subscriber's state type 
+    ///   implements `Equatable`. Defaults to `false`.
     public required init(
         reducer: @escaping Reducer<State>,
         state: State?,
-        middleware: [Middleware<State>] = []
+        middleware: [Middleware<State>] = [],
+        automaticallySkipsRepeats: Bool = false
```
mjarvis: The changes look fine to me, my only question would be whether we want to change the default on users.
It might be better to have the default be `true`, which maintains backwards compatibility with no change required by users, and yet allows opt-out for those who wish.

Ben-G:
I also wonder whether we should change the default behavior. If we think the new behavior is better we should. Then the next release just needs to get a major version bump.

Also @DivineDominion, would you mind adding a Changelog entry? Thanks a lot for the work on this 🎉

ReSwift/CoreTypes/Store.swift
```diff
@@ -40,11 +40,29 @@ open class Store<State: StateType>: StoreType {
 
     private var isDispatching = false
 
+    /// Indicates if new subscriptions attempt to apply `skipRepeats` 
+    /// by default.
+    let subscriptionsAutomaticallySkipRepeats: Bool
```
DivineDominion: Good point -- I really wanted to make this `public` but forgot. My rationale: if this is a major change in "modes", it might be useful to query for the currently active mode. 

But I see that since it's configured once and cannot change, you are supposed to know the mode anyway. So `private` makes more sense.

ReSwift/CoreTypes/Store.swift
DivineDominion: Makes sense.

DivineDominion:
Thinking about your concerns, changing the default may not be necessary and I'll revert that if you agree. So many people wondered why duplicate state is passed on. _Not_ skipping repeats is a more advanced setting. If you're that far into using ReSwift, you should be able to discover this parameter in the docs.

Speaking of the docs, I think it's a good idea to leave a note about this parameter in the guide, too.

- [ ] add info to docs
- [ ] make property private
- [ ] change default parameter value

DivineDominion:
Changes made ✔️

@Ben-G we could consider changing the default value to `false` in an upcoming release if this should turn out to be too much of a surprise for people in v5. I wonder how we could find out how people use this in the real world 🤔

Ben-G:
@DivineDominion Thanks for all the work on this 🎉!  Merge ahead.

DivineDominion:
@Ben-G  Done ✓ Looking at the commit log, I now wonder if you'd all favor squash & merge instead to produce less noise 🤔 If so, I'll add that to the `CONTRIBUTING` file for reference.

mjarvis:
I generally prefer squash and merge, unless the PR has important, distinct commits that are worth keeping separate.

Ben-G:
Agree! Sorry, forgot to specify that. Next time 😄 
Adding to `CONTRIBUTING` would be great @DivineDominion!

DivineDominion:
Did add something in #272  @Ben-G @mjarvis

PR #270, opened 2017-06-28
Remove StandardAction

### LINKED ISSUE #269: Should `StandardAction` be moved to ReSwift-Recorder?
It seems like the existence of `StandardAction` and `StandardActionConvertible` is somewhat confusing to new users, and does not really have a direct use within the base ReSwift library.

Without a compelling reason to leave it, I suggest we move it to `ReSwift-Recorder`.

mjarvis:
The existence of StandardAction and StandardActionConvertible is somewhat confusing to new users, and does not really have a direct use within the base ReSwift library.

Therefore, I am moving it to `ReSwift-Recorder`

Closes #269


mjarvis:
Yes I have a branch ready with ReSwift-Recorder, was waiting for this to be merged so I could update the version of ReSwift over there

PR #278, opened 2017-08-01
Fix retain cycle in SubscriptionBox

mjarvis:
`SubscriptionBox` retains `originalSubscription`, which was retaining `self` in the observe closure.

We can safely make this `[unowned self]` since `SubscriptionBox` is the only object that holds a reference to `originalSubscription` (it is discarded in the `Store` function that creates the box)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ matrix:
     - osx_image: xcode8.3
       env: SCHEME="macOS"  SDK="macosx10.12"          DESTINATION="arch=x86_64"
     - osx_image: xcode8.3
-      env: SCHEME="iOS"    SDK="iphonesimulator10.3"  DESTINATION="OS=10.3,name=iPhone 6S Plus"
+      env: SCHEME="iOS"    SDK="iphonesimulator10.3"  DESTINATION="OS=10.3.1,name=iPhone 6S Plus"
     - osx_image: xcode8.3
       env: SCHEME="tvOS"   SDK="appletvsimulator10.2"  DESTINATION="OS=10.2,name=Apple TV 1080p"
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,11 +16,11 @@ env:
 
 matrix:
   include:
-    - osx_image: xcode8.2
+    - osx_image: xcode8.3
       env: SCHEME="macOS"  SDK="macosx10.12"          DESTINATION="arch=x86_64"
-    - osx_image: xcode8.2
+    - osx_image: xcode8.3
       env: SCHEME="iOS"    SDK="iphonesimulator10.2"  DESTINATION="OS=10.1,name=iPhone 6S Plus"
-    - osx_image: xcode8.2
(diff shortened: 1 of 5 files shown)


mjarvis:
Somewhat related: While looking at this I was looking at the possibility of `transformedSubscription` causing a retain cycle as well but I can't even find where its being retained at all. It seems to me like after line 35 of `Subscription.swift` that `transformedSubscription` is no longer referenced, and the `subscribe` function of `Store` does not seem to retain it either (not referenced after the initialization of `SubscriptionBox`.

This would mean that the transformed subscription is being released immediately after creating, but that would then mean that transformed subscriptions would not function, no? I'm quite confused and feel like I'm missing something obvious.

DivineDominion:
@mjarvis I know how confused you feel :) I always need a while to re-orient myself in the codebase. And I only have pet theories and heuristics to explain to myself what's going on.

`SubscriptionBox.init`'s `transformedSubscription is indeed not kept as a direct reference there. The `.observe` call only sets up the insides of the box to be used when the subscription fires. (I hope that's an accurate description, anyway.) I do wonder if the call to `originalSubscription.observe` strongly captures `self` and thus will never let the `SubscriptionBox` go away. And ... well, I've got bad news, if I got my test right :)

Have a look at: 
https://github.com/ReSwift/ReSwift/blob/73f6d890dde683707461c0cbd3dfb95f1ab02633/ReSwiftTests/StoreSubscriptionTests.swift#L42

DivineDominion:
Breaking at the failing test and having a look at the _Memory Graph Hierarchy_ in Xcode reveals:

screen shot 2017-08-02 at 18 07 11

mjarvis:
So adding `[unowned self]` to `Subscription.swift:L35` resolves that cycle as well, but I'd like to make sure we understand where `transformedSubscription` is being retained, otherwise we're just guessing that breaking this cycle with `unowned` won't cause issues. If the only reference to it is being held inside `SubscriptionBox` somehow (down the chain is ok), then we're safe. But if something else is the only thing holding the transformed subscription (or its `observe` closure), then `unowned` might not be safe

mjarvis:
Oh I guess your image there confirms that we're safe -- looks like `SubscriptionBox` retains `originalSubscription` which retains the transformed subscription.

DivineDominion:
It appears so, but I don't quite get _how._ After all, we do retain 2 `Subscription` objects in the chain. But `Store.swift` at line 107 shows that the transformation is called on the original subscription, not the other way round. I'm still puzzled and wonder if this test now passing is a false indicator.

mjarvis:
I've been running this branch for a few days now, and the leaks are gone, and we've had no negative effects. Lets get a code-review on this so we can merge.

DivineDominion:
👍 It seems I can remove my test branch because it doesn't serve any purpose anymore. Will do in a couple of days if there's no veto.

mjarvis:
@DivineDominion Yes, you are safe to remove your test branch, as I merged your changes here.

PR #295, opened 2017-09-28
Update Swiftlint

mjarvis:
* Removes an unnecessary disable of `function_body_length`
* Resolves `multiple_closures_with_trailing_closure` warning
* Resolves a couple private vs fileprivate warnings


DivineDominion:
Just discovered this, ran it locally, and it seems to look fine ✓

PR #296, opened 2017-09-29
WIP Side Effect Only Middleware (e.g. Logging, Analytics)

mikerunnals:
This seems to be related to #188 but with more limited scope of a change.

### DIFF
--- a/ReSwift/CoreTypes/Middleware.swift
+++ b/ReSwift/CoreTypes/Middleware.swift
@@ -11,3 +11,21 @@ import Foundation
 public typealias DispatchFunction = (Action) -> Void
 public typealias Middleware<State> = (@escaping DispatchFunction, @escaping () -> State?)
     -> (@escaping DispatchFunction) -> DispatchFunction
+
+public typealias SideEffectOnlyMiddleware<State> = (Action, State?) -> Void
+
+public enum Middlewares {
+    case normal(Middleware<StateType>)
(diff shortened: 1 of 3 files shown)


DivineDominion:
Thanks for the PR @mikerunnals!  This would be quite a bit of a change to the API, so can you tell us more about your motivation behind this and where you want to go with this in the long term?

---

Because creating a `Middleware` for a simple side effect is so laborious, I am now experimenting with a `SideEffect` type that's like a simpler `Middleware`, too. Unlike your approach, it can be installed _as a middleware_ and forward actions. (As a downside, I'm using a singleton/global collection of `SideEffect`s.) I wrote about it here: <http://cleancocoa.com/posts/2017/09/reswift-middleware-different-target/>

As long as something similar would suffice, I think making changes to the ReSwift core would be overkill.

mikerunnals:
Thanks @DivineDominion, this is good info. I'll take a look...

This PR is really just for conversation. One aspect of it is that the current Middleware is unchanged an a new type is supported via a convenience initializer on the store that simply converts Simple Middleware to the normal/existing Middleware type. I think it could be done entirely in extensions. 

In the end if seems a lot of client code **repeats** the same few lines that are somewhat confusing to read:

```
let _ mw: Middleware<StateType> = {
    return { dispatch, getState in
        { next in
            { action in
                sideEffect(action, getState())
                next(action)
            }
        }
    }
}
```

Issue #298, opened 2017-10-26
Default skipRepeates when State conforms equatable only works for top level subscribers


danielmartinprieto:
Hi!

I've recently seen something in my app that I find weird, and I think it's because I'm not understanding how the default behaviour for subscriptions works when the state conforms equatable.

All my state is equatable, so when I create a subscription using `select { ... }`, I expect the subscriber to be notified only when the substate of the subscription has changed, but this seems not to be the case. If I add `skipRepeats()` after the select, then it works, but per the [documentation](https://github.com/ReSwift/ReSwift/blob/master/Docs/Getting%20Started%20Guide.md), concretely 

> For example, one requirement may be to notify subscribers only when the state they are interested in has changed. This is the default Store setting. 

I wasn't expecting that I should add `skipRepeats()` every time I use `select { ... }`.

I checked the [code](https://github.com/ReSwift/ReSwift/blob/master/ReSwift/CoreTypes/Store.swift#L198) and it seems that indeed it would work only if I subscribe to the top level (the whole state). See `open func subscribe<S: StoreSubscriber>(_ subscriber: S) where S.StoreSubscriberStateType == State`.

So I don't know if I don't understand the docs properly, or the code, or there is something else that I'm missing. Ideally, by default I should be able to subscribe to a substate automatically skipping repeats if my whole state conforms Equatable, right? And, actually, I would say this was like this before and it changed at some point.

Thanks!

danielmartinprieto:
@DivineDominion hey, do you know something about this?

DivineDominion:
Ok, so even though `state.subscriptionsAutomaticallySkipRepeats` is `true`, after you `select` a substate, you get identical results? That sounds not like what I'd expect, too. 

Of course I don't have all the code memorized :) Looking at [how the auto-skipping works](https://github.com/ReSwift/ReSwift/blob/master/ReSwift/CoreTypes/Store.swift#L198), `skipRepeats` seems to be called on the initial super-state, not the substate, which is kind of wasteful, too.

I'm on a very tight schedule right now and cannot spend as much time to work on this. If you could add some failing unit tests in a PR, we could move this issue forward together towards a solution.

danielmartinprieto:
Cool! I will put something together soon.

danielmartinprieto:
Done! https://github.com/ReSwift/ReSwift/pull/299

JoeCherry:
#300 Has a potential fix

PR #300, opened 2017-11-07
Fix for #298

JoeCherry:
related to #298 and #299


ReSwift/CoreTypes/Subscription.swift
```diff
@@ -80,6 +80,13 @@ public class Subscription<State> {
         return self._select(selector)
     }
 
+    public func select<Substate: Equatable>(
+        _ selector: @escaping (State) -> Substate
+        ) -> Subscription<Substate>
+    {
+        return self._select(selector).skipRepeats()
```
danielmartinprieto: I'm not sure if this is handled in another place, but shouldn't we check if `subscriptionsAutomaticallySkipRepeats` is true before skipping?

JoeCherry:
Instead of overriding the actual select function I overrode the subscribe function. I don't like duplicating all that code, but I'm not seeing another way to do it at the moment.

danielmartinprieto:
@JoeCherry Cool! A way of avoiding some duplication could be either to call the original function (if possible) from your new function, passing it the new transformation, or to extract the common parts of both functions in a private one so the core logic is only in one place.

JoeCherry:
Broke out some logic in the subscription functions to a separate function. I believe this solution should fix the issue.

ReSwift/CoreTypes/Store.swift
```diff
@@ -204,4 +212,17 @@ extension Store where State: Equatable {
             }
             _ = subscribe(subscriber, transform: { $0.skipRepeats() })
     }
+
+    open func subscribe<SelectedState: Equatable, S: StoreSubscriber>(
```
Ben-G: This looks pretty good. But wouldn't it be possible to implement this method as part of the extension (`extension Store where State: Equatable`)? That might be more in line with the existing implementation.

(I'm traveling right now, so can't come up with what the actual implementation would look like 😉 )

Ben-G:
My internet connection was better than expected, to I had some time to to look into this. I actually noticed some ambiguity in the documentation that's referred to in #298. The original intention was to only skip state updates, if the entire state remains unchanged.

The unit tests here indicate that @danielmartinprieto expects updates to also be skipped if the selected substate does not change (even if not selected state might have changed).

With that in mind, I think you can disregard my earlier comment on this diff, as we need to implement skipping based on the selected state, not the overall state of the store.

ReSwift/CoreTypes/Store.swift
```diff
@@ -85,40 +85,48 @@ open class Store<State: StateType>: StoreType {
...
-        // Call the optional transformation closure. This allows callers to modify
-        // the subscription, e.g. in order to subselect parts of the store's state.
-        let transformedSubscription = transform?(originalSubscription)
-
         let subscriptionBox = self.subscriptionBox(
-            originalSubscription: originalSubscription,
-            transformedSubscription: transformedSubscription,
+            originalSubscription: original,
+            transformedSubscription: transformed,
             subscriber: subscriber
```
Ben-G: I actually prefer the more explicit argument names. What do you think about reverting this part of the diff?

ReSwift/CoreTypes/Store.swift
Ben-G: See comment below; this can be disregarded.

ReSwift/CoreTypes/Store.swift
JoeCherry: No problem

ReSwiftTests/TestFakes.swift
```diff
@@ -31,6 +31,22 @@ extension TestStringAppState: Equatable {
...
 
+struct TestNonEquatable: StateType {
+    var testValue: NonEquatable
+
+    init() {
+        testValue = NonEquatable()
+    }
+}
+
+struct NonEquatable: StateType {
```
mjarvis: this shouldn't need to be `StateType` since its used as a substate. `StateType` is only necessary for top-level state

ReSwift/CoreTypes/Store.swift
```diff
@@ -204,4 +213,18 @@ extension Store where State: Equatable {
...
+
+    open func subscribe<SelectedState: Equatable, S: StoreSubscriber>(
+        _ subscriber: S, transform: ((Subscription<State>) -> Subscription<SelectedState>)?
+        ) where S.StoreSubscriberStateType == SelectedState
+    {
+        let originalSubscription = Subscription<State>()
+
+        var transformedSubscription = transform?(originalSubscription)
+        if subscriptionsAutomaticallySkipRepeats {
+            transformedSubscription = transformedSubscription?.skipRepeats()
```
mjarvis: What will happen here now if we call `skipRepeats()` manually on substate with this enabled? Will it cause a double-skip-repeats? (Sort of relates to https://github.com/ReSwift/ReSwift/pull/259 )

Aside: I think the amount of subscription bugs & things we've been dealing with lately are calling for a more substantial refactor of the internals of subscriptions, given the amount of side-effects that are showing up

JoeCherry: Right now, as I understand it, it will cause a double skip repeats which is the same thing outlined in #259.

mjarvis: Alright, just wanted to confirm that. I'm okay with it having the same behaviour as we already do now.

mjarvis:
Thanks @JoeCherry! Like all contributors I'll invite you to the ReSwift org. That will make future contributions easier in case you are interested! 🎉

I'd also like to invite you to check out https://github.com/ReSwift/ReSwift/pull/307 where I'm taking a more core-refactor look at the Subscription type, fixing this and other things (the duplicate calls, cleaner implementation, more extensible). I think your input would be valuable there (be honest! if you hate it thats okay.)

mjarvis:
I'm going to update the changelog in here before merging this -- please give me some time I'm currently caught in another set of work

mjarvis:
actually, scratch that, I can't since its in a fork.
@JoeCherry can you please add this to the changelog?

`Add automatic skipRepeats for Equatable substate selection (#300) - @JoeCherry`

JoeCherry:
@mjarvis Do you want me to merge this in to master? I'm still seeing a merge conflict even though master was just merged in. Also, we can probably close #299 as I have those test cases in this PR.

mjarvis:
Weird. I used github's UI to resolve the conflicts, I guess it didn't take the first time?
Anyways, I'll merge this in the morning after the CI completes

danielmartinprieto:
Awesome! Thanks, @JoeCherry!

PR #305, opened 2017-11-23
Readme swift4

fabiothiroki:
Since Swift3, some UIKit methods had the first parameter label removed from the method signature, so I've updated the example contained in the `README.md` to be consistent with this change.


DivineDominion:
Excellent, thank you! I'll hereby ask @mjarvis to add you to the org :)

Issue #263, opened 2017-06-20
Improving the Subscription transformation API


DivineDominion:
Thinking back to #203, @Ben-G wasn't super happy with the nested closures, like:

    store.subscribe(foo) { $0.skipRepeats().select { $0.substate } }

We're now passing in some kind of Builder object, using `Subscription`. But what if we expose the building?

### Introducing a source stream

    store.stateUpdates
        .skipRepeats()
        .select { $0.substate } 
        .subscribe(foo)

I imagine `subscribe` to return `Void` and the other operators/transformations requiring the client to do something with the intermediate result. Thus the compiler will nag the programmer to finish the building process.

### Keeping `subscribe` at the front

Closer to the existing API:

    store.subscribe(foo)
        .skipRepeats()
        .select { $0.substate } 
        .XXX()

The "XXX" will be tricky, though; "commit" might work, but "subscribe" implies the side effect already. A pair of "prepare(foo) / subscribe()" is less elegant, too.

Ben-G:
These are interesting ideas! Probably the first approach is preferable because it logically ends in a subscribe. I'm not entirely sure this change is worth a breaking API change? But in case you have a working prototype it would be nice to be able to try it out!

PR #307, opened 2017-11-27
Proof of concept inverting subscription transforms

mjarvis:
The intention of this PR is to simplify subscription selection & skipping, as well as resolve a number of bugs & negative performance implications of using `skipRepeats` in the current implementation.

Fixes:
- https://github.com/ReSwift/ReSwift/issues/298
- https://github.com/ReSwift/ReSwift/issues/263

```swift
// Previously this:
store.subscribe(subscriber) {
    $0.select { $0.testValue }
        .skip { $0 == $1 }
}
// Now is this:
store.subscription()
    .select { $0.testValue }
    .skip { $0 == 1 }
    .subscribe(subscriber)
```

TODO:
- [ ] Filter & remove subscriptions without subscribers (need to walk to the end)
- [ ] automaticallySkipRepeats (`extension StoreSubscriber where StoreSubscriberStateType: Equatable` doesn't work / isn't called when the storesubscriber itself is generic, `class Bar<S>: StoreSubscriber`)
- [ ] unsubscribe
- [ ] retain cycle tests

### DIFF
--- a/ReSwiftTests/StoreSubscriberTests.swift
+++ b/ReSwiftTests/StoreSubscriberTests.swift
@@ -82,6 +82,9 @@ class StoreSubscriberTests: XCTestCase {
 
     // TODO: Add tests for splitting after `select`, `skip`, and `only`
 
+    // TODO: Add test for subscribing and splitting (I dont think this works right now)
+    // eg: `let sub = subscription(); sub.subscribe(subscriber); sub.select({ $0.xyz }).subscribe(otherSubscriber)
+
     /**
      it does not notify subscriber for unchanged substate state when using `skipRepeats`.
      */


mjarvis:
My latest commit here refactors `Subscription`, breaking backwards compatibility.
The intention here is to make the Subscription system much easier to understand, code-wise, eliminating a number of confusing code paths & bugs.

Subscriptions now look like this:
```
store.subscription()
    .select { $0.testValue }
    .skip { $0 == 1 }
    .subscribe(subscriber)
```

ReSwift/CoreTypes/Store.swift
```diff
@@ -17,17 +17,17 @@ import Foundation
...
     // TODO: Setter should not be public; need way for store enhancers to modify appState anyway
     // swiftlint:enable todo
 
     /*private (set)*/ public var state: State! {
         didSet {
-            subscriptions = subscriptions.filter { $0.subscriber != nil }
+//            subscriptions = subscriptions.filter { $0.subscriber != nil }
             subscriptions.forEach {
-                $0.newValues(oldState: oldValue, newState: state)
+                $0.notify(oldValue, state)
```
DivineDominion: Love this 👍

ReSwift/CoreTypes/Store.swift
```diff
@@ -85,51 +85,13 @@ open class Store<State: StateType>: StoreType {
...
-        ) -> SubscriptionBox<State> {
-
-        return SubscriptionBox(
-            originalSubscription: originalSubscription,
-            transformedSubscription: transformedSubscription,
-            subscriber: subscriber
+    open func subscription() -> Subscription<State> {
+        let subscription = Subscription<State>(
+            initialState: self.state,
+            automaticallySkipsEquatable: self.subscriptionsAutomaticallySkipRepeats
```
DivineDominion: Why not use the same parameter name, one way or the other? If this is a breaking API change anyway, I’m d’accord with renaming automaticallySkipsRepeats.

ReSwift/CoreTypes/Store.swift
```diff
@@ -85,51 +85,13 @@ open class Store<State: StateType>: StoreType {
...
         )
+        self.subscriptions.append(subscription)
```
DivineDominion: This means that store.subscription() already inserts a subscription, even though subscribe() is never called, right?

If so, I think we should change that to operate on a pending subscription that is only activated in the store on subscribe(). This is now more like a factory; the side effect can cause troubles. If you do not call append here, you do not run into problems when you grab a handle of the subscription during its transformation and intend to branch off in 2+ directions. That’d be similar to RxSwift.Observalbe stream transformations and its subscribe call, which is a prominent and rather well understood pattern.

ReSwift/CoreTypes/Subscription.swift
```diff
@@ -8,159 +8,87 @@
...
+        }
+        return subscription
     }
-}
 
-extension Subscription where State: Equatable {
-    public func skipRepeats() -> Subscription<State>{
-        return self.skipRepeats(==)
+    public func subscribe<S: StoreSubscriber>(_ subscriber: S) where S.StoreSubscriberStateType == State {
+        assert(self.subscriber == nil, "Subscriptions do not support multiple subscribers")
```
DivineDominion: See comment above; not allowing multiple calls to subscribe() is unintuitive and makes the function a partially applied function, if I remember the term correctly, through the (debug build-only) assertion. I think this should be a precondition that affects client code, because if not, the behavior is supposed to be undefined.
 but will not show any sign of a problem.

That being said, in accordance to the stuff above, I think it should be subscribe(store:) instead, with a PendingSubscription from store.subscription() being sugar to hold on to a Store and Subscription instance, and PendingSubscription.subscribe() simply passing the store to the subscription.

DivineDominion:
I really like the overall changes! That reads much nicer than the old block stuff.

In general, I think the side effect of activating the subscription in the Store.subscription() factory problematic. I left a few comments with suggestions inside. I dunno how much time you want to spend on this; If you want me to propose a change in another branch for comparison instead of leaving the implementation to you, I’ll happily come up with code changes, just say Yes :)

ReSwift/CoreTypes/Store.swift
mjarvis: I'm for changing that as well, I think `subscriptionsAutomaticallySkipEquatable` is much more clear as to what its doing.

mjarvis:
@DivineDominion I've tried to see if I can get anywhere with your suggestions but I keep running into walls, or the implementation gets confusing enough that its not worth it. I've put up my attempt in https://github.com/ReSwift/ReSwift/tree/mjarvis/invert-transform-2. My commit description has a bunch of notes / known issues etc. I would love for you to do a bit on this and see if you can simplify things

mjarvis:
@DivineDominion I've managed to partially implement tree-style subscription (allowing multiple `select/skip/only/subscribe` on a given `Subscription`) (I put a couple `TODO` in the tests)
I think that covers _most_ of your concerns above. (maybe? 😄)

DivineDominion:
@mjarvis Aha! So the store becomes a hub of subscriptions that each manages its own subscribers, right? Sounds clever, and reasonable. The concepts work, kind of: a subscription has many subscribers, and a store can have many ongoing subscriptions. (Maybe calling the intermediate products “subscription” conveys too much of a successful, finalized binding? What about StateUpdate for the intermediary objects, with the result of the final subscribe() returning a Subscription?) Have to check out the code on my Mac and run it tomorrow to see everything in context, esp. if a subscription with 0 subscribers removes itself from the store again and stuff like that :) What I can make sense of on GitHub looks good so far!

DivineDominion:
I experimented with the changes and tried to fix some broken tests, then I split the Subscription type into subtypes to see if that makes getting a handle on things easier than the closure dance from before because then things have names in the stack trace. It kind of does, I find, but I have yet to fix any of the operators leaving a strong reference. Your approach was clever, with the notify/originalNotify being captured in the next operator’s closure. I’d like to reverse that, though, and will need to spend some more time on this tomorrow. 

Will see if I can manage to push the changes right now, which I forgot to do back at my desk :)

DivineDominion:
@mjarvis here it is so far: https://github.com/ReSwift/ReSwift/tree/DivineDominion/invert-transform-types

PR #312, opened 2017-12-15
[Docs] Describe Release Process

Ben-G:
When I'm not working, I don't want to be the person blocking releases from rolling out.

Per @mjarvis's good suggestion I'm documenting the lightweight release process. I'm also giving @mjarvis and @DivineDominion CocoaPods ownership access, such that more folks can publish new versions of ReSwift.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -131,3 +131,16 @@ Here's a checklist for you about what makes Pull Requests to ReSwift shine:
 * You can propose PRs to merge with the `master` branch directly. We don't use any complex branching strategies.
 
 As a contributor, choose the "squash & merge" strategy to merge PRs with a single commit, keeping the commit history clean. (That's an upside of focused Pull Requests: you don't lose extra information.)
+
+## Publishing New Releases
+
+Members of the [@ReSwift/releases](https://github.com/orgs/ReSwift/teams/releases) team have the necessary permissions to publish a new version to CocoaPods. If you want a new version of ReSwift to be published you should ping these folks.
+
+To create a new release, create a pull request targeting either `master`, or a separate release branch for hot-fixes, with the following:
+
+- [ ] Bump version number in `ReSwift.podspec` and `Info.plist` (We follow [semver](https://semver.org/) - most importantly any breaking API change should result in a major API version bump)
+- [ ] Add the new version number and the release date to `Changelog.md`
+- [ ] Create a tag off of the relevant branch (`master` for regular release) and add the relevant changelog entries to the [release list on GitHub](https://github.com/ReSwift/ReSwift/releases) 
+- [ ] Publish the new version to the CocoaPods trunk
+- [ ] ✨✨Bonus points ✨✨: for major version`ReSwift` releases, update `ReSwiftRouter` and `GitHubBrowserExample` to use the latest version of `ReSwift`
+


CONTRIBUTING.md
```diff
@@ -131,3 +131,25 @@ Here's a checklist for you about what makes Pull Requests to ReSwift shine:
...
+
+To create a new release:
+
+- [ ] Bump version number in `ReSwift.podspec` and `Info.plist` (We follow [semver](https://semver.org/) - most importantly any breaking API change should result in a major API version bump)
+
+- [ ] Add the new version number and the release date to `Changelog.md`
+
+  **Depending on release type:**
+  - [ ] If you're releasing off of `master`, simply create a PR against `master` after the version has been bumped
+  - [ ] If you're creating a release with cherry-picked fixes, create a PR against a separate release branch
```
mjarvis: I don't understand what is referring to for PR. Is this in reference to a PR that updates the version number & release date as mentioned above? If so, perhaps this should be worded just below `To create a new release`, we say something along the lines of "Create a pull request targeting either `master`, or a separate release branch for hot-fixes, with the following:"

CONTRIBUTING.md
```diff
@@ -131,3 +131,25 @@ Here's a checklist for you about what makes Pull Requests to ReSwift shine:
...
+
+- [ ] Create a tag off of the relevant branch (`master` for regular release)
```
mjarvis: We should specify that the tag should be created via https://github.com/ReSwift/ReSwift/releases with the changelog release notes copied over

CONTRIBUTING.md
```diff
@@ -131,3 +131,25 @@ Here's a checklist for you about what makes Pull Requests to ReSwift shine:
 * You can propose PRs to merge with the `master` branch directly. We don't use any complex branching strategies.
 
 As a contributor, choose the "squash & merge" strategy to merge PRs with a single commit, keeping the commit history clean. (That's an upside of focused Pull Requests: you don't lose extra information.)
+
+## Publishing New Releases
+
+Currently @DivineDiminion, @mjarvis, @Ben-G and @agentk are folks that have the necessary permissions to publish a new version to CocoaPods. If you want a new version of ReSwift to be published you should ping these folks.
```
mjarvis: I've added a new GitHub team for the organization named @ReSwift/releases -- That way if the team of people capable to release changes we don't need to update this document.

DivineDominion: Good idea 👍 Having a link to the team would be easier to maintain than editing the GitHub handles of users with the access rights. Also, I think we should recruit a 5th person because I'm not the best candidate for this as I don't even work with CocoaPods and have far less experience with all this infrastructure :)

mjarvis: @DivineDominion No worries. If you're not comfortable with this, we're all happy with your current level of contributions. We're just saying that we trust you to have the permissions to do so if you chose to.

My desire is just making sure that we have enough people with permissions that, if a quick hotfix is needed, theres someone that can step up and do so without waiting weeks.

CONTRIBUTING.md
Ben-G: Used your wording 👍

CONTRIBUTING.md
Ben-G: Great idea, will link to the team instead!

Ben-G:
@mjarvis  updated with your suggestions, thanks for the swift review!

Ready for re-review.

PR #325, opened 2018-02-22
Improve subscription & state update performance

### LINKED ISSUE #324: Slow subscribe ?
Hi,

I've a project with thousands of subscribers, the more I add the slower it becomes to add new ones.
I think the main reason is here

```
        if let index = subscriptions.index(where: { $0.subscriber === subscriber }) {
            subscriptions.remove(at: index)
        }
```

unfortunately this is going to be a major issue for my project; would you consider a PR for a different approach like an hashtable or whatever ? 

thanks in advance

mjarvis:
Fixes #324

See commit descriptions for details

TL;DR:
- testNotify(): 0.0046359 -> 0.0031482 (32% faster)
- testSubscribe(): 0.1861 -> 0.0087853 (95% faster)

### DIFF
--- a/ReSwift/CoreTypes/Store.swift
+++ b/ReSwift/CoreTypes/Store.swift
@@ -25,10 +25,15 @@ open class Store<State: StateType>: StoreType {
 
     /*private (set)*/ public var state: State! {
         didSet {
-            subscriptions = subscriptions.filter { $0.subscriber != nil }
+            var filtered = subscriptions
             subscriptions.forEach {
-                $0.newValues(oldState: oldValue, newState: state)
+                if $0.subscriber == nil {
+                    filtered.remove($0)
+                } else {
+                    $0.newValues(oldState: oldValue, newState: state)
+                }
             }
+            subscriptions = filtered
         }
     }
 
--- a/ReSwift/CoreTypes/Store.swift
+++ b/ReSwift/CoreTypes/Store.swift
@@ -36,7 +36,7 @@ open class Store<State: StateType>: StoreType {
 
     private var reducer: Reducer<State>
 
-    var subscriptions: [SubscriptionType] = []
+    var subscriptions: Set<SubscriptionType> = []
 
     private var isDispatching = false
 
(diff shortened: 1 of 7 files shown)


ReSwift/CoreTypes/Store.swift
```diff
@@ -25,18 +25,23 @@ open class Store<State: StateType>: StoreType {
 
     /*private (set)*/ public var state: State! {
         didSet {
-            subscriptions = subscriptions.filter { $0.subscriber != nil }
+            var filtered = subscriptions
             subscriptions.forEach {
```
JoeCherry: I think this could be written as:
```swift
subscriptions = subscriptions.filter {
    if $0.subscriber != nil {
        $0.newValues(oldState: oldValue, newState: state)
        return true
    }
    return false
}
```
instead of assigning a variable, filtering, and then reassigning the variable

mjarvis: I try to avoid side-effects in "pure" functions such as `filter`, `map`, etc. I will, however, test this out to see if its faster, as if theres some magic in `filter` that improves the performance its probably worth it.

IanHoar:
This looks great! The performance gains are exciting!

ReSwift/CoreTypes/Store.swift
DivineDominion: I too like to put side-effects outside, but from that point of view, I actually prefer the old `.filter` & `.forEach` sequence to the newly proposed approach which looks oddly imperative next to the rest of the code base. But the old code has a complexity of O(2n) in the worst case, while the new imperative code and @JoeCherry's side-effect-ful filter should be O(n), only, am I right?

My bet is on `.filter` with side effect being faster, because `Set.remove` takes additional time (the docs say it's an amortized complexity of O(1), so would this turn @mjarvis approach into O(n+1)? 🤔)

mjarvis: Performance tests show no recordable difference between this PR and @JoeCherry's `filter` side-effect suggestion.

mjarvis: I think I'm happy merging this, as its gotten a few eyes on it to make sure I haven't caused any regressions. Just need an approved review from someone :)

JoeCherry: Cool, it looks good the way it is

Issue #304, opened 2017-11-14
Enable - Allow app extension API only


amreshkumar:
ReSwift should be usable across widgets and iMessage targets. To be able to do that "Allow app extension API only" should be enabled.

amreshkumar:
When used with cocoapod this setting can be set in Pods project. But is there a way to resolve this while using carthage?

DivineDominion:
I recently discovered this when checking the library target's build settings for RxSwift. I have 0 experience with app extensions; can you try flicking the build setting and see what happens? This should directly affect Carthage (while CocoaPods basically provides a build settings override which can become redundant this way, too).

PR #328, opened 2018-03-12
(iOS only) Enable build settings "Allow app extension API only"

oradyvan:
Since I am working on an iOS app extension, I was planning to re-use part of our main app code that is using ReSwift library. To avoid the warnings about unsafe use of non-app extension API in the app extension I tried enable the build setting "Allow app extension API only" in the project (both Debug and Release configurations). Nothing else has changed, except for the fact I am now able to use ReSwift in any app extensions for iOS platform.

Here is the official list of APIs that cannot be used by ReSwift framework with these build settings enabled:
https://developer.apple.com/library/content/documentation/General/Conceptual/ExtensibilityPG/ExtensionOverview.html#//apple_ref/doc/uid/TP40014214-CH2-SW6
As for me they all look really non-related to the functionality ReSwift framework provides so I expect no issues with that.

Here is related open issue for this change: https://github.com/ReSwift/ReSwift/issues/304

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Add Subscription `skip(when:)` and `only(when:)` (#242) - @mjarvis
 - Add `automaticallySkipsRepeats` configuration option to Store initializer (#262) - @DivineDominion
 - Improve subscription & state update performance (#325) - @mjarvis
+- Enable build settings "Allow app extension API only" (#328) - @oradyvan
 
 # 4.0.1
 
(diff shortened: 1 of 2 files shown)


mjarvis:
Thanks @oradyvan! Like all contributors I'll invite you to the ReSwift org. That will make future contributions easier in case you are interested! 🎉

mjarvis:
@oradyvan Before I can merge this, can you please add this change to `CHANGELOG.md` under `Other:`.

`- Enable build settings "Allow app extension API only" (#328) - @oradyvan`

PR #330, opened 2018-03-21
Fix disappearing subscription bug from #325

mjarvis:
If a new subscriber was added in `newState` caused by a state change,
assigning `subscriptions = filtered` would then discard that new
subscription.

Since `subscriptions` is a value type, we're safe to mutate in the
`forEach`, so we can fix this bug by never holding a copy of subscriptions,
instead just mutating in-place. If we changed `subscriptions` to a
reference type we would need to enumerate a copy instead.

Performance tests match baselines within allowable margins for error.

### DIFF
--- a/ReSwift.xcodeproj/xcshareddata/xcbaselines/25DBCF861C30C4DB00D63A58.xcbaseline/Info.plist
+++ b/ReSwift.xcodeproj/xcshareddata/xcbaselines/25DBCF861C30C4DB00D63A58.xcbaseline/Info.plist
@@ -28,6 +28,30 @@
 			<key>targetArchitecture</key>
 			<string>x86_64</string>
 		</dict>
+		<key>CA98BC4B-4A03-4B07-BF9E-12C3F6A52195</key>
+		<dict>
+			<key>localComputer</key>
+			<dict>
+				<key>busSpeedInMHz</key>
+				<integer>100</integer>
+				<key>cpuCount</key>
+				<integer>1</integer>
+				<key>cpuKind</key>
+				<string>Intel Core i7</string>
+				<key>cpuSpeedInMHz</key>
+				<integer>2800</integer>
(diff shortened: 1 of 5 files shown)


ReSwift.xcodeproj/xcshareddata/xcbaselines/25DBCF861C30C4DB00D63A58.xcbaseline/Info.plist
```diff
@@ -28,6 +28,30 @@
...
+				<string>MacBookPro14,3</string>
+				<key>physicalCPUCoresPerPackage</key>
+				<integer>4</integer>
+				<key>platformIdentifier</key>
+				<string>com.apple.platform.macosx</string>
+			</dict>
+			<key>targetArchitecture</key>
+			<string>x86_64</string>
+		</dict>
 	</dict>
```
Ben-G: Seems like performance test artifacts slipped in by mistake?

IanHoar:
I can confirm that this fixes a number of instances where subscribers were not being notified in my project. Nice work!

ReSwift.xcodeproj/xcshareddata/xcbaselines/25DBCF861C30C4DB00D63A58.xcbaseline/Info.plist
mjarvis: The performance test artifacts are necessary for testing changes to the codebase, as the baselines are needed to see if there are differences in performance. Anyone with the same computer as the baselines set will get a test failure if they regress performance.
I like these being committed as a historical record of where the test was performed & what the result was.

mjarvis:
To note: I've intentionally omitted updating the changelog since this bug fix is for a change that exists only on master at the moment. If it had been released, I'd include this PR in the changes.

PR #334, opened 2018-04-01
Added Redux-Twitter example to README.md

Goktug:
I developed a basic Twitter search application built with ReSwift and RxSwift. I think it's a really good real-world example.


Ben-G:
Thanks a bunch ⭐️🎉
Great to have more examples!

I will also add you as a contributor to the ReSwift org in case you want to contribute more in future!

PR #341, opened 2018-05-02
Make isDispatching of Store atomic

zhongwuzw:
1. Make isDispatching of Store atomic.
2. Add `Xcode9.3` IDEWorkspaceChecks.plist.
3. Fixed typo.


ReSwift/Utils/AtomicBool.swift
```diff
@@ -0,0 +1,26 @@
...
+//  Created by 钟武 on 2018/5/2.
+//  Copyright © 2018年 Benjamin Encz. All rights reserved.
+//
+
+import Foundation
+
+/**
+ Struct is only used internally in ReSwift to implements atomic bool operation.
+ */
+internal struct AtomicBool {
```
DivineDominion: When you make this conform to `ExpressibleByBooleanLiteral`, you can avoid accessing the `.value` property in `Store.swift` and compare with (or assign to) `true`/`false` directly

DivineDominion:
Interesting! Did you have problems in the past with the existing approach?

ReSwift is supposed to be run from a single thread, so I wonder if there's any upside to wrapping this in atomic get/set calls (which I am totally not familiar with :))

ReSwift/Utils/AtomicBool.swift
zhongwuzw: @DivineDominion , Emm, `ExpressibleByBooleanLiteral` only support initialized with the `Boolean` literals. So we can't compare directly. 
Maybe we can just use `.value` property to access, it's use for internal.

zhongwuzw:
@DivineDominion It's the latent issue.

> ReSwift is supposed to be run from a single thread

Yep, we haven't force users to call on single thread, so it still exists concurrency, but we just can't dispatch the same time.
`OSAtomicTestAndSet` is `lock-free` I think, I just run the performance test, and it's almost the same compared the original `Bool` implementation.

ReSwift/Utils/AtomicBool.swift
DivineDominion: Quite the contrary, == is defined for ExpressibleByBooleanLiteral, too:
https://gist.github.com/DivineDominion/12b96f625cb640dd88d4f0ebdf5f59fe

zhongwuzw: We can't use this method as you posted. every time when we set `isDispatching`, we do make a new struct, it still not thread safe.

DivineDominion:
@zhongwuzw Sorry for the long silence! I think I now better understand the implementation. I recently discovered RxSwift to use a similar lock for its internal flags. Can you rebase your PR and add your PR to the Changelog?

DivineDominion:
@zhongwuzw The tests apparently are not linked against a framework that exposes the lock. Can you fix that?

zhongwuzw:
@DivineDominion I think I'm done.

DivineDominion:
Sorry for the delay, and many thanks for the contribution!

basememara:
We can use property wrappers for this which is also backwards compatible with <iOS13: https://www.onswiftwings.com/posts/atomic-property-wrapper/

Also I think semaphore is the way to go as long as its not on the main thread. `OSAtomicTestAndSet` feels like magic maybe because I don't know much about it but many are saying its deprecated. GCD seems the way to go since it more Swifty and even faster and safer than kernel calls because its dynamic based on environment and conditions. Just not sure how supported GCS is on Linux if ReSwift consumed there.

```
@Atomic var counter: Int = 1
--------
@propertyWrapper
struct Atomic<Value> {

    private var value: Value
    private let lock = NSLock()

    init(wrappedValue value: Value) {
        self.value = value
    }

    var wrappedValue: Value {
      get { return load() }
      set { store(newValue: newValue) }
    }

    func load() -> Value {
        lock.lock()
        defer { lock.unlock() }
        return value
    }

    mutating func store(newValue: Value) {
        lock.lock()
        defer { lock.unlock() }
        value = newValue
    }
}
```

DivineDominion:
@basememara Sounds interesting! Could you open a new PR for this? I have had 0 experience with the new Swift features so far, btw, and would be happy for others to chime in and provide feedback :)

basememara:
Turns out property wrappers cannot handle concurrency without Swift implementing coroutines. I wrote about it here: https://basememara.com/creating-thread-safe-generic-values-in-swift/

I also found these kernel level locks are the fastest, but drain the CPU. It's always better to leverage GCD so it can manage resources properly (if Linux doesn't have to be supported). This is ok for now, but the scenario that concerns me is when the store is hammered for long periods of time, otherwise the CPU drain is negligible.

Issue #264, opened 2017-06-22
make Store.state setter private


DivineDominion:
Found this little gem in the `Store` code today:

     // TODO: Setter should not be public; need way for store enhancers to modify appState anyway
    /*private (set)*/ public var state: State!

Making the setter private doesn't break any tests. 

@Ben-G added this back in commit 0effbd7  (Jan 2016) -- What kind of enhancers did you have in mind?

mjarvis:
@Ben-G Been a while. Do you recall what kind of enhancers you had in mind here?

Ben-G:
Hey folks! Sorry, missed the notification here. The only current user of this API would be ReSwift-Recorder. It needs the ability to rewind/forward to specific states without dispatching actions. 
Since I basically have done no work on the recorder since the original prototype I think we can go ahead and fix this TODO. If I or anyone else puts effort into building a non-prototype recorder we can figure out a way to extend the store API to accommodate that.

DivineDominion:
I'm not against breaking the API if it's all just for _ReSwift-Recorder_, but I think some people are using it in production 🤔

Was there ever any reason not to make the recorder come with a state replacement reducer and dispatch an action that triggers this change? If not, maybe we can push that first, then make the state private again.

danielmartinprieto:
Using ReSwift-Recorder or the setter?

DivineDominion:
Using the Recorder. Breaking the setter wouldn't be too much of a problem for a major version release, but not offering a working Recorder might be a downer for some folks.

danielmartinprieto:
I understand. To be honest I've never seen the code of the Recorder library. As far as I know, a recorder is usually implemented as a higher order reducer, which means that setting the state directly shouldn't be necessary, but if it's implemented like that, it's true that it'd need a rewrite.

More importantly, being able to set the state directly in ReSwift could be confusing for people that are new to the library, because it's an anti-pattern and, depending on the implementation of the library, it could cause bugs on the client side, so I'd remove it as soon as possible.

mokagio:
> More importantly, being able to set the state directly in ReSwift could be confusing for people that are new to the library, because it's an anti-pattern and, depending on the implementation of the library, it could cause bugs on the client side, so I'd remove it as soon as possible.

The ability to read and write the `StateType` from `store.state` has lead to all sorts of misuses of ReSwift in our codebase 😞 .

mokagio:
[...]

PR #354, opened 2018-06-05
Make state an internal property of Store

mokagio:
This PR removes the `State` property from the `StoreType` protocol, and makes it `private(set) internal` in the `Store` implementation.

The aim of this change is to promoted a better usage of ReSwift, preventing consumers from falling into the anti-pattern of reading the `State` directly from the `Store`, rather than conforming to `StoreSubscriber`. See the discussion on #264.


ReSwift/CoreTypes/StoreType.swift
```diff
@@ -18,9 +18,6 @@ public protocol StoreType: DispatchingStoreType {
 
     associatedtype State: StateType
 
-    /// The current state stored in the store.
-    var state: State! { get }
```
mjarvis: I don't think this change should be included in this PR. There are valid reasons for making `state` unreadable, but there are also valid reasons for reading state outside of a subscriber (IE one-time reads). Whether or not it stays I think this should be discussed & dealt with separately from the change to removing write access.

Dimillian: I agree with that. Making the state totally private from the store could be a source of a lot of problem in some architectures.

danielmartinprieto: The original Redux implementation in JavaScript has a `getState()` method to query the state of the store at any time, so that should be enough, IMO!

DivineDominion: Instead of Java-style getters like `getState()`, making the property puplicly readable via `public private(set)` could work.

Dimillian: Would be perfect. Setter can and should be private anyway.

danielmartinprieto: Yeah, they probably do it because of JavaScript, but in Swift we could just use the read-only variable. It's the same as in the middleware, you get both the `dispatch` function and `getState`.

mokagio:
Hey folks. Thanks for all the great comments. Unfortunately I haven't had and still don't have the bandwidth to follow this up 😞 .

Feel free to pick up the branch if you want, or close this PR and start a new one.

PR #356, opened 2018-06-10
Remove unnecessary imports

dduan:
Most files doesn't use symbols from `Foundation`, the imports,
therefore, are unnecessary.


mjarvis:
Thanks @dduan ! Like all contributors I'll invite you to the ReSwift org. That will make future contributions easier in case you are interested! 🎉

SD10:
Regarding the above @mjarvis I merged the PR for ReSwift/reswift-peril causing this error. I was waiting for review and then it slipped my mind

PR #362, opened 2018-07-04
Util for Middlewares

Tavernari:
My motivation to create this utils is about the how difficult is to create a middleware.

### Example to implement

```
import Foundation
import ReSwift

class Logger:MiddlewareExecutor<State>{
    override func execute(action: Action, state: State?, nextDispatcher: @escaping DispatchFunction) -> Action? {
        print("[\(String(describing: action))] called\n")
        return action
    }    
}
```

_If return `nil` on executor the queue of middleware will stop_

```
let appMiddlewares = MiddlewaresCollection<State>()
appMiddlewares.add(Logger())
appMiddlewares.add(OtherMiddleware())
            
store = Store<State>(reducer: appReducer, state: State(), middleware: appMiddlewares.middlewares);
```

### DIFF
--- a/ReSwift/Utils/MiddlewaresCollection.swift
+++ b/ReSwift/Utils/MiddlewaresCollection.swift
@@ -3,22 +3,25 @@ import ReSwift
 
 class MiddlewareExecutor<T: StateType> {
     func execute(action: Action, state: T?, nextDispatcher: @escaping DispatchFunction) -> Action?{
-        return nil
+        return action
     }
 }
 
 class MiddlewaresCollection<T: StateType> {
     private var _middlewares: [Middleware<T>]
+    
     init(){
         self._middlewares = []
     }
     
     func concact(withCollection: MiddlewaresCollection) -> MiddlewaresCollection {
         self._middlewares = _middlewares + withCollection.middlewares
+        
         return self
     }
     
     func add(_ middlewareItens: MiddlewareExecutor<T>...) -> MiddlewaresCollection {
+        
         for item in middlewareItens {
             self._middlewares.append ({ (_, state) -> (@escaping DispatchFunction) -> DispatchFunction in
                 return { next in
--- a/ReSwift/Utils/MiddlewaresCollection.swift
+++ b/ReSwift/Utils/MiddlewaresCollection.swift
@@ -20,7 +20,7 @@ class MiddlewaresCollection<T:StateType>{
     
     func add(_ middlewareItens:MiddlewareExecutor<T>...)->MiddlewaresCollection{
         for item in middlewareItens {
-            self._middlewares.append ({ (dispatch, state) -> (@escaping DispatchFunction) -> DispatchFunction in
+            self._middlewares.append ({ (_, state) -> (@escaping DispatchFunction) -> DispatchFunction in
                 return { next in
                     return { action in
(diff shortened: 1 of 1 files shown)


DivineDominion:
This is an interesting addition, @Tavernari! Why did you choose to make the `execute` function return an action? That's a deviation from middlewares in two ways:

- you can replace the current action with a new one if you return it (which is just another way of `dispatch`ing a different action and not `next`)
- you cannot "next" more than 1 action (I have not used next multiple times, but I'm not 100% sure if that would be useful)

DivineDominion:
I think one of the problems is the nested closures not working well with Xcode auto completion. :/

danielmartinprieto:
I didn't have any bad experience with this in the past, so I can't help you guys there. As this _util_ is something built on top on the current primitives (and they don't change the primitives themselves) I wouldn't include it in this library.

Maybe if you still need it you can create a micro framework?

Tavernari:
> This is an interesting addition, @Tavernari! Thanks for contributing!
> 
> I wonder: Why did you choose to make the `execute` function return an action? What are your reasons for this change? (And should we maybe change the Middleware to be similar if your reasons are sound?) That's a deviation from middlewares in two ways:
> 
> * you can replace the current action with a new one if you return it (which is just another way of `dispatch`ing a different action and not `next`)
> * you cannot "next" more than 1 action (I have not used next multiple times, but I'm not 100% sure if that would be useful)

Hi @DivineDominion thank you for reply.

About the action return 

    func add(_ middlewareItens:MiddlewareExecutor<T>...)->MiddlewaresCollection{
        for item in middlewareItens {
            self._middlewares.append ({ (dispatch, state) -> (@escaping DispatchFunction) -> DispatchFunction in
                return { next in
                    return { action in
                        if let nextAction = item.execute(action: action, state: state(), nextDispatcher: next){
                            next(nextAction);
                        }
                    }
                }
            });
        }
    
        return self;
    }

I wrapped the implementation in this method and for don't break the flow we need to call `next` function so I am returning the action, in this case you can interrupt the flow only returning `nil` in your `execute` method.

If I make some mistake when I implemented the `middleware` plase let me know, I like so much this framework.

Tavernari:
> I think one of the problems is the nested closures not working well with Xcode auto completion. :/

Exactly

> I didn't have any bad experience with this in the past, so I can't help you guys there. As this _util_ is something built on top on the current primitives (and they don't change the primitives themselves) I wouldn't include it in this library.
> 
> Maybe if you still need it you can create a micro framework?

My intention isn't to change the main implementation but only add a way to help newbies on lib to use middleware, so it isn't necessarily need be implemented in this library, I think it can be a Plugin to ReSwift to help others to do this well. 

What do you think if I publish it as a Plugin separated?

DivineDominion:
Thanks for explaining! @danielmartinprieto's comment is spot-on: if it is a convenience wrapper, it's better to publish it as an extension instead of inside the ReSwift lib.

I think there were other middleware-related code snippets flying around in the issues here. These might be interesting to create a compilation of useful middleware tools, for example.

Tavernari:
@DivineDominion @danielmartinprieto 

I created a micro-library as how you indicate and this URL is https://cocoapods.org/pods/ReSwiftMiddleware

It is possible can you check if it is ok and If it has some change to do?

DivineDominion:
Added tavernari/ReSwiftMiddleware#1 👍

PR #365, opened 2018-07-20
Copy-Edit Getting Started Guide

jnozzi:
* Edited wording for clarity and readability.
* Grammar corrections.
* Fixed errors in and updated code examples.
* Added image caption for vision impaired.

### DIFF
--- a/Docs/Getting Started Guide.md
+++ b/Docs/Getting Started Guide.md
@@ -2,21 +2,19 @@
 
 ReSwift provides the infrastructure for `Store`s, `Action`s and `Reducer`s to enable a unidirectional data flow as shown below.
 
-![](img/reswift_detail.png)
+![Detailed flowchart of data flow with ReSwift](img/reswift_detail.png)
 
-The following section will briefly describe how to set up the core components of ReSwift in your app.
+The following section will briefly describe how to set up the core components of ReSwift in your app. The next section will describe the components that power ReSwift in depth.
 
-The section afterwards will describe the components that power ReSwift in depth.
+To see ReSwift in action, [have a look at the repositories on GitHub](https://github.com/ReSwift/), which contains example apps and playgrounds.
 
-To see ReSwift in action, [have a look at the repositories on GitHub](https://github.com/ReSwift/) to download example apps and playgrounds.
+# General Setup and High-Level Overview
 
-# General Setup and High Level Overview
-
-This section will show you how to set up the core ReSwift components in your app quickly. It is intended to provide a quick start to get the ball rolling.
+This section will show you how to set up the core ReSwift components in your app. It is intended to provide a quick start to get the ball rolling.
 
 ## High-Level Overview
 
-To get the infrastructure up and running, you need to set up a `Store` for a root app state type. The type requirement is `ReSwiftStore<State: ReSwift.StateType>`. You set it up like this:
+To get the infrastructure up and running you need to set up a `Store` for a root app state type. The type requirement is `ReSwift.Store<State: ReSwift.StateType>`. You set it up like this:
 
 ```swift
 import ReSwift
@@ -35,12 +33,12 @@ let store = Store(
     middleware: [])      // Middlewares are optional
 ```
 
-Keep the store around somewhere (for example in an `NSDocumentController`, the `AppDelegate`, or as a global variable). 
+Keep the store around somewhere. For example in an `NSDocumentController`, the `AppDelegate`, or as a global variable.
 
 The process of implementing then works like this:
 
 1. You can add **subscribers** to the store and **dispatch actions**. 
-2. The actions will be pre-processed by your **middleware**. The middleware can act on actions and pass them along, producing side effects or dispatch additional actions itself. It can also do some side effects like logging to the console and pass the action along. 
+2. The actions will be pre-processed by your **middleware**. The middleware can act on actions and pass them along, producing side effects or dispatching additional actions itself. It can also do some side effects like logging to the console and passing the action along. 
 3. Finally, if any action was passed through all middleware, it'll reach the root **reducer**, here called `appReducer` by convention. The reducer changes the **state** according to the incoming action. 
 4. The resulting state is then stored in the **store**. The store consequently propagates the new state to all **subscriptions**, reaching the **subscriber** objects if the subscription requirements are met. For example, one requirement may be to notify subscribers only when the state they are interested in has changed. This is the default `Store` setting. If you want to pass on identical states after a reducer pass, have a look at `automaticallySkipsRepeats` in the `Store`'s initializer.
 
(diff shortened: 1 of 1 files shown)


Docs/Getting Started Guide.md
```diff
@@ -152,28 +140,29 @@ struct SetOAuthURL: Action {
 
 ## Reducers
 
-Reducers are the only place in which you should modify application state! Reducers take the current application state and an action and return the new transformed application state. We recommend to provide many small reducers that each handle a subset of your application state.
+Reducers are the **only place** in which you should **modify** application state. Reducers take the current application state and an action then return the new transformed application state. We recommend to provide many small reducers that each handle a subset of your application state.
 
-You can do this implementing a top-level reducer that conforms to the `Reducer` protocol. This reducer will then call individual functions for each different part of the app state.
+You can do this by implementing a top-level reducer that conforms to the `Reducer` protocol. This reducer will then call individual reducer functions for each different part of the app state, according to their areas of responsibility.
```
DivineDominion: Excellent changes so far! While we're at it, there's no protocol anymore, just a typealias for the function signature. You may want to change the wording accordingly.

Docs/Getting Started Guide.md
```diff
@@ -152,28 +140,29 @@ struct SetOAuthURL: Action {
...
 
-Here's what the `authenticationReducer` function that is called from the `appReducer` looks like:
+The `Reducer` typealias is a function that takes an `Action` and a `State?` then returns a `State`. Typically reducers will be responsible for initializing the application state. When they receive `nil` as the current state, they should return the initial default value for their portion of the state. In the example above the `appReducer` delegates all calls to other reducer functions. 
+
+For example, `authenticationReducer` is responsible for providing the `authenticationState`. Here's what the `authenticationReducer` function might look like:
 
 ```swift
-func authenticationReducer(state: AuthenticationState?, action: Action) -> AuthenticationState {
+func authenticationReducer(_ state: AuthenticationState?, _ action: Action) -> AuthenticationState {
```
DivineDominion: To match the call above (Line 152), this would have to be `authenticationReducer(_ action: Action, _ state: AuthenticationState)`. I prefer named parameter, to be frank, though.

Docs/Getting Started Guide.md
```diff
@@ -81,7 +77,7 @@ struct AppState: StateType {
 // This action does not have state and is a mere marker of "X happened":
 struct AddAction: Action { }
 
-func appReducer(action: Action, state: AppState?) -> AppState {
+func appReducer(_ action: Action, _ state: AppState?) -> AppState {
```
DivineDominion: Line 150 is `appReducer(action: Action, state: State?)` with named parameters, making this inconsistent. I personally prefer named parameters, especially since the function name isn't revealing what the 2 parameters would be, unlike, say `add(x, y)`.

Docs/Getting Started Guide.md
```diff
@@ -279,10 +270,10 @@ store.subscribe(self) { subcription in
 
 ### Asynchronous Operations
 
-Conceptually asynchronous operations can simply be treated as state updates that occur at a later point in time. Here's a simple example of how to tie an asynchronous network request to `ReSwift` state update:
+Conceptually asynchronous operations can simply be treated as state updates that occur at a later point in time. Here's a simple example of how to tie an asynchronous network request to a `ReSwift` state update:
 
 ```swift
-func fetchGitHubRepositories(state: State, store: Store<State>) -> Action? {
+func fetchGitHubRepositories(_ state: State, _ store: Store<State>) -> Action? {
```
DivineDominion: Aside from the reducer signatures, what's your motivation to remove parameter names here?

Docs/Getting Started Guide.md
```diff
@@ -320,24 +311,36 @@ func fetchGitHubRepositories(state: State, store: Store<State>) -> Action? {
 }
 ```
 
-In the example above, we're using an `enum` to represent the different states of a single state slice that depends on a network request (e.g. loading, result available, network request failed). There are many different ways to model states of a network request but it will mostly involve using multiple dispatched actions at different stages of your network requests.
+In the example above we're using an `enum` to represent the different states of a single state slice that depends on a network request. It may look something like this:
+
+```swift
+enum GitHubFetchRequestState {
+    case Loading
```
DivineDominion: To comply with Swift naming conventions, I suggest lowercasing the cases to `case loading`, `case repositories`, `case networkError`.

Docs/Getting Started Guide.md
```diff
@@ -320,24 +311,36 @@ func fetchGitHubRepositories(state: State, store: Store<State>) -> Action? {
...
 ```swift
-typealias ActionCreator = (state: State, store: StoreType) -> Action?
+typealias ActionCreator = (_ state: State, _ store: StoreType) -> Action?
 ```
 
 A very simple example of an `ActionCreator` might be:
+
 ```swift
-func doubleValueIfSmall(state: TestAppState, store: Store<TestAppState>) -> Action? {
+func doubleValueIfSmall(_ state: TestAppState, _ store: Store<TestAppState>) -> Action? {
```
DivineDominion: Same as above -- even though the typealias cannot provide parameter names anymore, the implementation may, and thus add context on call sites.

jnozzi:
I’m confused about the method signatures. When I found the initial error of swapped arguments in the last two appReducer calls to sub-reducers, I looked up the signature in the repository and confirmed the correct order and also noticed they were unnamed (_), so I reflected that in the corrections. Was I looking in the wrong place?

DivineDominion:
Ah, I see: it's just that Swift (4?) enforces name-less signatures when you define blocks and typealiases. That's why you have to write `(_ foo: Foo) -> Bar` both for typealias definitions and block parameters in functions. This essentially conveys: "I don't care about the parameter name, only about the type". That's the intent of the change in Swift, I think; _But_ you can pass in functions with fully qualified parameter names as long as the order of parameter types is the same, so `func baz(aName: Foo) -> Bar` will satisfy `(Foo) -> Bar` and `(_ foo: Foo) -> Bar`. So you write functions as Swift-y as before.

jnozzi:
Alright! Sorry for the delay (I got a bit busy). I've amended the PR to include the requested changes. Without specific guidance on the signature matching recommendation portion, I went ahead and added a note explaining the discrepancy and changed "matches" to "resembles". I also changed the protocol references for Reducer to reflect the current state of the repo. I did see other protocol mentions are still protocols in code, so I left those alone.

jnozzi:
Any updates on review? Any more changes needed?

mjarvis:
Thanks @jnozzi ! Like all contributors I'll invite you to the ReSwift org. That will make future contributions easier in case you are interested! 🎉

jnozzi:
Branch deleted and org invite accepted, thanks!

DivineDominion:
Sorry for the silence last workweek! Good job 👏

PR #369, opened 2018-08-01
Refactor newState() to apply(state:)

pepasflo:
Thanks so much for writing ReSwift!

Due to my personal baggage from languages of the past, a function named `newState()` is hard-wired in my brain to mean something like "Create and return a new state object".

This PR simply renames `newState()` to `apply(state:)`.

Feel free to merge if you think this is an improvement.  If not, no hard feelings 😁

### DIFF
--- a/Docs/Getting Started Guide.md
+++ b/Docs/Getting Started Guide.md
@@ -200,11 +200,11 @@ After the `appReducer` has called all of the sub-reducer functions, we have a ne
 
 ## Store Subscribers
 
-Store subscribers are types that are interested in receiving state updates from a store. Whenever the store updates its state it will notify all subscribers by calling the `newState` method on them. Subscribers need to conform to the `StoreSubscriber` protocol:
+Store subscribers are types that are interested in receiving state updates from a store. Whenever the store updates its state it will notify all subscribers by calling the `apply(state:)` method on them. Subscribers need to conform to the `StoreSubscriber` protocol:
 
 ```swift
 protocol StoreSubscriber {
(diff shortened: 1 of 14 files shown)


pepasflo:
Oops, didn't intend to trigger pod-related changes.

DivineDominion:
I actually like how this reads way better than `newState`.

@mjarvis what do you think, maybe for 5.0 as a breaking API change, and as a deprecation for `4.1.0`?

DivineDominion:
@pepasflo could you undo the CocoaPods related changes for the scope of this PR, and mark `newState` as deprecated without removing it? (`@available(*, deprecated: 1.0, renamed: apply(state:))` or so)

pepasflo:
Sure! 😃

(might not get to it until this evening)

DivineDominion:
@mjarvis We never discussed this anywhere: I like the new wording, but what do you think?

danielmartinprieto:
I agree with @mjarvis. I’m not saying the current function name is perfect, but `apply` seems something more imperative than reactive, I didn’t say anything because I thought you already made the decision, but if one of the main function names of the framework were to be changed, I would spect a big list of alternatives so we can discuss and pick the best one.

brownsoo:
I also think the current function name is not perfect, but OK. I think `onNewState` seems more appropriate than `apply` in condition to be changed the name. because the function is called reactively. (^^; a minor opinion)

pepasflo:
I'll go ahead and close this for now.  The seed has been planted :)

danielmartinprieto:
Hahaha. Fair enough.

Issue #371, opened 2018-09-04
Documentation on AsyncActionCreator


sohocoke:
What's the best resource to learn about AsyncActionCreator and when to use it? 

It's not mentioned in the excellent Getting Started Guide, and the comment on the typealias:
```
/// AsyncActionCreators allow the developer to wait for the completion of an async action.
```
still leaves me with questions such as:
- how exactly is this different from an ActionCreator? (since an ActionCreator can return nil, and 'wait until the completion of an async operation' until it dispatches actions)
- what are the situations for which it is useful?

danielmartinprieto:
Hi! This is something that has come up several times already (take a look at the issues). I find them misleading as well because they don't represent they same concept as in the original Redux library (in JavaScript), so I don't use them myself. If I want to include side effects in my app, I follow one of these patterns:
- Crete a middleware (preferred)
- Create functions that get the store and possibly other params and dispatch new actions

I hope this helps.

sohocoke:
@danielmartinprieto Thanks for your response! 

Unfortunately my search didn't turn up any issue that explained what makes AsyncActionCreator unique and where I would use them... Could you kindly point me to the ones you were thinking of?

danielmartinprieto:
Sure, you can find them [here](https://github.com/ReSwift/ReSwift/issues?utf8=%E2%9C%93&q=is%3Aissue+AsyncActionCreator).

DivineDominion:
Would be great if you add your learnings to the Getting Started guide in a PR!

danielmartinprieto:
Let me know what you think about [this PR](https://github.com/ReSwift/ReSwift/pull/372) regarding this!

PR #372, opened 2018-09-07
ReSwift thunks

danielmartinprieto:
In order to get rid of the confusion created by the _action creators_ and _async action creators_, discussed several times here, I've created this PR with a proposal:
- I've removed the code related with _action creators_. As some people have mentioned before in the issues of this repo, an action creator in Swift is just the constructor of the type conforming to the `Action` protocol. This is different from the original Redux written in JavaScript; because the actions there are just plain objects, they created these pure functions in order to avoid mistakes assigning the correct `type` key (a string) and making them easier to create. Of course, in Swift we can also have a function that returns an instance or value of a type conforming `Action`, and then dispatch it, but unless the construction of that action is super complex, I don't think it's justified and, at the end, what the store really cares about is just an `Action`, independently of the way it was created.
- I've also removed the code related with _async action creators_, which I think they are even more confusing, in conjunction with the previously mentioned _action creators_, because you can do the same with both of them and it's not clear where's the distinction. Again, the store should only be able to dispatch plain actions, and the only place where side effects can happen should be the middleware chain. This makes reasoning about ReSwift a lot easier, reducing the API surface of the library.
- One way of producing side effects (there are several) in a simple manner and widely used by the original Redux library users is called _thunks_. The way this is implemented in JavaScript is through a little external package called [`redux-thunk`](https://github.com/reduxjs/redux-thunk), which gives you just a way of creating a middleware that will check every action and treat the _special_ ones in a certain way, letting the rest pass through the middleware chain, unaltered. I've implemented the Swift version of this middleware, adapting it the way ReSwift already works. Note that this code could perfectly live in an external package as well, given that it's just an extension and doesn't need access to private parts of the current library, or modify it in any way. This is an example of how it works:
```swift
// First, you create the middleware, which needs to know the type of your `State`.
let thunksMiddleware: Middleware<MyState> = createThunksMiddleware()

// Note that it can perfectly live with other middleware in the chain.
let store = Store<MyState>(reducer: reducer, state: nil, middleware: [thunksMiddleware])

// A thunk represents an action that can perform side effects, access the current state of the store, and dispatch new actions, as if it were a ReSwift middleware.
let thunk = Thunk<MyState> { dispatch, getState in 
    if getState!.loading {
        return
    }
    dispatch(RequestStart())
    api.getSomething() { something in
        if something != nil {
            dispatch(RequestSuccess(something))
        } else {
            dispatch(RequestError())
        }
    }
}

// As the thunk type conforms to the `Action` protocol, you can dispatch it as usual, without having to implement an overload of the `dispatch` function inside the ReSwift library.
store.dispatch(thunk)

// Note that this action won't reach the reducers, instead, the thunks middleware will catch it and execute its body, producing the desired side effects.
```

I hope you like this change, as it simplifies the ReSwift API and its usage, and brings it closer to how the original Redux library works. I don't use thunks myself, neither in JavaScript nor Swift. Instead, I create a middleware for each entity that produces side effects in my program (an API module, a music player, a local database, etc.), and dispatch plain actions with the purpose of reaching one of those middleware, the reducer, or both. This is why I think this thunks middleware and the `Thunk` type could be considered an extension of ReSwift and exist on their own package.

I've added some tests as well, but I haven't touched the docs, as I wanted to get feedback from the community first. I can later add some comments and docs, and maybe also put examples of how I avoid thunks just using specialised middleware.

Finally, as a technical note, I've implemented the `Thunk` type as a struct conforming to `Action`. At the beginning, I was trying to accomplish the same declaring a `Thunk` protocol extending `Action`, but it was cumbersome dealing with PATs and found the struct approach simpler. Let me know if you have ideas to improve both the API and its implementation!


mjarvis:
I like the simplification of removing action creators & async action creators, though I think we have to be careful removing functionality. If we are to remove them, they should be first deprecated for a major release before being entirely removed.
I do think that the `Thunk` functionality should be separate from the ReSwift core.
Perhaps if enough people think it is an important tool, it could be a sub-spec? IE `ReSwift/Thunk`.

sohocoke:
As the newbie who first raised the question of the purpose of AsyncActionCreator (#371), I think the bit of this PR that decommissions AsyncActionCreator is probably a good answer to the question, if all the elders can agree it was indeed a redundant / unnecessary abstraction.

However the part of getting rid of ActionCreators, I’m not too sure of. 

- From my limited exposure, the ‘reference’ basic technique to create side effects in Redux seems to be by writing action creators. Thunks and other middleware-based solutions might be better for many cases, but do raise the initial hurdle of the number of concepts to fully grasp before a user can do something useful. I don’t see much harm in having action creators around, as long as there is enough help to so the user can ‘graduate to’ something else for side-effecting. (this help seems a little thin btw — middleware usage examples are very scarce)

- ActionCreators can be dispatched to the store. An action creator has logic on what to do in order to create actions, but we do not execute it; we cede control to the store. With no ActionCreators, any similar implementation of async work will have to be executed explicitly — they can’t just be given to the store. If this needs to be done by the view controller, it means the neat pattern of simply sending stuff from the view layer to the store is eroded a little. The answer could again be some middleware abstraction, but I’m not sure whether I would always want async activity to be so far from my view controllers, even through I’m definitely sure the control not being in the view controllers is a good thing.

As an side, from the point of view of understanding things coming from the need to write swift apps, I found the easiest way to understand Actions and ActionCreators was simply to think of them as Dispatchables. So, provided the clarity justifies deviating from Redux jargon, we can just have it so that the store can accept`Dispatchable`s, to which `Action` or `() -> Action?` conforms?

danielmartinprieto:
@mjarvis IMO, releasing a major version and describing the breaking changes should be ok, right? People can still use the current version and move on when they're ready. About `ReSwift/Thunk`, if we ship it with the main library, it will help people discover it and learn how to use the middleware chain, and they will also have a way of producing side effects built in (as it is now with the "action creators" and "async action creators"). I prefer moving it to a new package, but that'll make discover it harder. If you create the new repo, I can move there the code, if you want.

@sohocoke the problem is that they both ("action creators" and "async action creators") have always caused confusion during these years because the concept was taken from the original Redux library but it wasn't _translated_ properly to the Swift world, where the rules are a bit different given the language idiosyncrasy. It will be easier to understand for newcomers that the only thing that can be dispatched is an `Action`, and the problems you arise regarding your view controllers having too much to do can be solved in multiple ways, depending on the architecture you are more comfortable with. For instance, you can create an entity (or just plain functions), pass them the store and let them do the async work for your view controller, for instance:
```
class MyViewController: UIViewController {
    let store: MyStore
    let myAPI: MyAPI

    viewDidLoad() {
        super.viewDidLoad()
        store.subscribe(self) { ... } // Subscribe to the store as you normally would
        getFriends(with: store, api: myAPI) // This function will do the async work for your view controller
        // ...
    }
}

// Note that this can be in another file, and even belongs to a certain class,
// it's up to you and how you like organising things.
func getFriends(with store: MyStore, api: MyAPI) {
    store.dispatch(GetFriendsStart())
    api.getFriends { error, friends in
        guard error == nil else { 
            store.dispatch(GetFriendsError(error))
            return
        }
        store.dispatch(GetFriendsSuccess(friends))
    }
}
```

sohocoke:
May I propose we de-complect the 3 separate topics in play:
- Introducing thunks
- Deciding what to do with Async Action Creators
- Deciding what to do with Action Creators

The code changes in the PR are to do with all 3, but it seems they are practically orthogonal to each other, and my comments below only pertain to the 3rd topic.

That aside, the `#getFriends` method is basically an action creator that dispatches the first action directly, rather than returning it. I can't see why that way of implementing async actions would be preferable to writing an action creator -- it basically seems to exactly the same thing, except not having the concept of an action creator would mean a missing abstraction in the framework. 

Perhaps it's because I'm unaware of the history of questions, but I myself was never confused about action creators given the clear Getting Started example -- it was only async action creators I didn't understand.

@danielmartinprieto Maybe the problem is confusion arising from different abstractions between Redux and ReSwift with the same name. Could you share some examples of the differences between Redux and ReSwift action creators? The ones I skimmed through didn't really strike me as all that different. If this is the case then I'd prefer to see the ReSwift ActionCreator renamed to something else rather than disappear, since I think it's a useful abstraction.

kedio:
@danielmartinprieto you said in your first post:
>  I can later add some comments and docs, and maybe also put examples of how I avoid thunks just using specialised middleware.

I would really like to see this example of how to avoid thunks by using specialized middleware. I've been going over the issues and doc on this repository and still can't find a consensus for doing async network calls.

I feel like the best approach is by using middleware and I like the idea to have specialized middleware by API/service instead of one thunk middleware that can be used for all sorts of side effects. I will probably have a shot at implementing it myself, but I would appreciate seeing how you do it!

danielmartinprieto:
@kedio Sure, it's pretty simple and you should try to implement it yourself, I'm sure you will come up with nice abstractions that fit your use case and you way of programming. For instance, you could do something like this:
```swift
// Note that if `MyAPI` is a protocol, the middleware gets super easy to test.
func createAPIMiddleware(with api: MyAPI) -> Middleware<AppState> {
    return { dispatch, getState in
        return { next in
            return { action in
                next(action) // Let the action pass through, in case other middleware/reducers are interested as well.
                switch action {
                case let _ as GetSomething: // This is the action we care about here to trigger the API call.
                    dispatch(GetSomethingStart()) // We dispatch a new action to tell our store that we're starting the API call.
                    api.getSomething { result in
                        switch result { // When the API call has finished, we dispatch actions with the result
                            case .success(let something):
                                dispatch(GetSomethingSuccess(something))
                            case .failure(let error):
                                dispatch(GetSomethingFailure(error))
                        }
                    }
                default:
                    break
                }
            }
        }
    }
}
```
And that's it, it's quite simple. Note how you could make these actions generic over the requests and responses types, reducing the boilerplate you have to write to interact with your API.

Following this pattern, you can create a middleware for analytics, for a local database, navigation, auth and all kind of side effects.

kedio:
@danielmartinprieto thank you for the example! I think it is the best approach to side effect I could find while all the ActionCreator/AsyncActionCreator stuff gets sorted out!

danielmartinprieto:
You’re welcome.

Even when it does, I truly believe using the middleware is the way to go to perform side effects using ReSwift. As I mentioned, I don’t use thunks myself, although I understand why some people find them useful.

ReSwift/CoreTypes/Thunk.swift
```diff
@@ -0,0 +1,33 @@
...
+public struct Thunk<State: StateType>: Action {
+    let body: (_ dispatch: @escaping DispatchFunction, _ getState: @escaping () -> State?) -> Void
+    public init(body: @escaping (
+                _ dispatch: @escaping DispatchFunction,
+                _ getState: @escaping () -> State?) -> Void) {
+        self.body = body
+    }
+}
+
+public func createThunksMiddleware<State: StateType>() -> Middleware<State> {
```
DivineDominion: Instead of this verbose free function, what do you think about an extension to Store like `Store.thunkMiddleware() -> Middleware<State>`? Then the factory would be contained in the main type people interact with; and the State type would be known. Alternatively, `Thunk<State>.middleware()` is not a bad place, either, in my opinion. What do you think, @mjarvis @danielmartinprieto ?

DivineDominion:
Nice work, @danielmartinprieto! 👏  I think this is a nice replacement and a good way to pave the future API of ReSwift. I second @mjarvis opinion on deprecating and merging the changes in a 4.x release soon, first, though. We could drop the ActionCreators in a 5.0 release completely then.

I'd also favor a `ReSwift-Thunk` µframework people can opt in to use. The 4.x deprecation message could then point to `https://github.com/ReSwift/ReSwift-Thunk/`. If someone wants me to create the repository for them, just ping me real quick.

ReSwift/CoreTypes/Thunk.swift
danielmartinprieto: I tried the second option you suggest and finally went with the free function because I didn't want to mess with the `Thunk` type, as the middleware would latter consume thunks, and this could be misleading for people. I also prefer not to put the static function in the main type because, as we've seen, this is a completely opt-in utility. I like the free function because it's simple, and it will always be namespaced by the framework, anyway, but let me know if you still prefer one of the other options or even create a new type for it (overkill, IMO).

danielmartinprieto:
@DivineDominion cool! if you can create it and give me access, I can leave in this PR the changes to remove what we don't need and point the docs to the new repo.

ReSwift/CoreTypes/Thunk.swift
DivineDominion: It sure is an opt-in when the new module is used. I wouldn't mind extending the core types in useful ways. I would prefer the factory to be an extension to `Middleware`, but sadly that's not possible :) Free functions in Swift feel weird to me, but I don't have any good arguments 🤷‍♂️

danielmartinprieto:
Thanks! I will try to put everything together soon.

ReSwift/CoreTypes/Thunk.swift
kedio: I also find free functions feels weird in Swift, I was wondering if it was like this because of the JS/Redux heritage or an actual benefit for ReSwift. I ended up creating a Reducer and Middleware protocol that wraps the free function in my example app.

danielmartinprieto: I didn't understand what you said about the reducer and the middleware protocol that wraps the free function. Could you elaborate?

kedio: I use this in my code:
```
protocol Middleware {
    associatedtype State

    func handle(action: Action, with getState: (() -> State?)?, and dispatch: DispatchFunction?)
}

extension Middleware {
    func create() -> ReSwift.Middleware<State> {
        return { dispatch, getState in
            return { next in
                return { action in
                    next(action)
                    self.handle(action: action, with: getState, and: dispatch)
                }
            }
        }
    }
}
```
So it can be used like this:
```
struct AssistantMiddleware: Middleware {
    // MARK: Constants

    typealias State = AppState

    // MARK: Properties

    var assistant: AssistantService = WatsonAssistantService.shared

    // MARK: Actions handling

    func handle(action: Action, with getState: (() -> AppState?)?, and dispatch: DispatchFunction?) {
        switch action {
        case ChatView.Actions.connect:
            assistant.send(message: nil, with: dispatch)

        case let InputBarView.Actions.send(message):
            assistant.send(message: message, with: dispatch)

        default:
            break
        }
    }
}
```
And for the reducer 
```
protocol Reducer {
    associatedtype StateType

    func reduce(_ action: Action, _ state: StateType?) -> StateType
}
```
So it can be used like this:
```
struct AppReducer: Reducer {
    // MARK: Constants

    typealias StateType = AppState

    // MARK: Properties

    let chatReducer = ChatReducer()

    // MARK: Actions

    func reduce(_ action: Action, _ state: AppState?) -> AppState {
        return AppState(
            chatState: chatReducer.reduce(action, state?.chatState)
        )
    }
}
```
Then I can just create my store like this:
```
final class AppStore: Store<AppState> {
    // MARK: Singleton

    static let shared = AppStore()

    // MARK: Initialization
    
    private convenience init() {
        let appReducer = AppReducer()
        let middlewares = [
            AssistantMiddleware()
        ]
        
        self.init(reducer: appReducer.reduce, state: AppState.defaults, middleware: middlewares.map { $0.create() })
    }

    internal required init(reducer: @escaping ReSwift.Reducer<State>, state: State?, middleware: [ReSwift.Middleware<State>]) {
        super.init(reducer: reducer, state: state, middleware: middleware)
    }
}
```
I try to stay away as much as possible from free funcution and global variable. I prefer to encapsulate everthing in a struct or class and have a singleton in case I want to access some common instance easily. 
Using this my app delegate is empty and when I want to access the store I do AppStore.shared.

DivineDominion:
@danielmartinprieto If you want me to handle the boilerplate (README, LICENSE, etc.) I can do that tomorrow on a public holiday

DivineDominion:
@danielmartinprieto I think I set things up similarly enough to the ReSwift main repo, including CI bots. That's my first time, so 🤞

DivineDominion:
Closed via https://github.com/ReSwift/ReSwift-Thunk/pull/2#

PR #374, opened 2018-09-21
Xcode 10.0 & Swift 4.2

mjarvis:
Migrates developer project to use Swift 4.2, Update CI configuration to add additional Xcode 10 test and updates iphone & apple tv tests to xcode 10.
This PR includes zero code changes, the project was already compatible with Swift 4.2, and thus no changelog entry is required.

### DIFF
--- a/ReSwift.xcodeproj/project.pbxproj
+++ b/ReSwift.xcodeproj/project.pbxproj
@@ -532,8 +532,6 @@
 		625E667A1C1FF97E0027C288 /* Project object */ = {
 			isa = PBXProject;
 			attributes = {
-				LastSwiftUpdateCheck = 0820;
-				LastUpgradeCheck = 0930;
 				ORGANIZATIONNAME = "Benjamin Encz";
 				TargetAttributes = {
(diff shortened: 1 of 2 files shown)


ReSwift.xcodeproj/project.pbxproj
```diff
@@ -547,11 +547,11 @@
 					};
 					25DBCF7A1C30C4AA00D63A58 = {
 						CreatedOnToolsVersion = 7.2;
-						LastSwiftMigration = 0900;
+						LastSwiftMigration = "";
```
DivineDominion: could we remove these lines completely?

mjarvis: I think we can somewhat remove them... I've now removed them and then ran a clean & build on each scheme, then resolved any warnings, and I also removed the target specifiers for swift version, so everything falls back to the project setting.

DivineDominion:
Interesting how the `LastSwiftMigration` setting behaves. But anyway, LGTM :) 👍

PR #380, opened 2018-11-10
Fix CocoaPods Link Directly

TaeJoongYoon:
I've changed link for CocoaPods so anybody can get to the link right away.


TaeJoongYoon:
Okay. Thank you😊

PR #383, opened 2018-11-20
Open Subscription to allow external extensions

mjarvis:
Makes `observer` and `init(sink:)` publicly accessible on `Subscription<State>`.
This allows for ReSwift users to extend subscription to add additional skipping or mutation functionality.

I've also removed `func observe` and changed the code to directly assign to `observer` to make the closure ownership more clear.

### DIFF
--- a/ReSwift/CoreTypes/Subscription.swift
+++ b/ReSwift/CoreTypes/Subscription.swift
@@ -36,13 +36,13 @@ class SubscriptionBox<State>: Hashable {
         // If we received a transformed subscription, we subscribe to that subscription
         // and forward all new values to the subscriber.
         if let transformedSubscription = transformedSubscription {
-            transformedSubscription.observe { [unowned self] _, newState in
+            transformedSubscription.observer = { [unowned self] _, newState in
                 self.subscriber?._newState(state: newState as Any)
             }
         // If we haven't received a transformed subscription, we forward all values
         // from the original subscription.
         } else {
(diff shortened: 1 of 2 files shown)


mjarvis:
For example, this would allow an application to add an extension that unwraps optional state, notifying only when non-nil:
```swift
public protocol OptionalProtocol {
    associatedtype Wrapped
    var value: Wrapped? { get }
}

extension Optional: OptionalProtocol {
    public var value: Wrapped? { return self }
}

extension Subscription where State: OptionalProtocol {
    public func skipNil() -> Subscription<State.Wrapped> {
        return Subscription<State.Wrapped> { sink in
            self.observe { oldState, newState in
                guard let newValue = newState.value else {
                    return
                }
                sink(oldState?.value, newValue)
            }
        }
    }
}
```

ReSwift/CoreTypes/Subscription.swift
```diff
@@ -36,13 +36,13 @@ class SubscriptionBox<State>: Hashable {
         // If we received a transformed subscription, we subscribe to that subscription
         // and forward all new values to the subscriber.
         if let transformedSubscription = transformedSubscription {
-            transformedSubscription.observe { [unowned self] _, newState in
+            transformedSubscription.observer = { [unowned self] _, newState in
```
DivineDominion: I like the change of names! 👍

PR #384, opened 2018-11-21
add watchOS build to Travis

DivineDominion:
Previously, Travis CI didn't cover all supported ReSwift platforms.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,15 +17,17 @@ env:
 matrix:
   include:
     - osx_image: xcode9
-      env: SCHEME="macOS"  SDK="macosx10.13"          DESTINATION="arch=x86_64"                    SWIFT_VERSION="3.2"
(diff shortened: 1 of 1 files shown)


DivineDominion:
@mjarvis  At work, we had unexpected problems with building for different targets, so I maybe got a bit overzealous :) Do you have experience with picking Travis SDKs? I wonder why we used simulator, when we could'be used the production target.

mjarvis:
@DivineDominion I've learned Xcode auto-selects the `simulator` variant when building to a simulator even if you select the device-oriented SDK, so the extra builds are actually exactly the same.
You can confirm this by looking at the logs for the device SDK travis builds, and note they're using `Debug-iphonesimulator` as output.
So we can pick either the simulator or the device variant to specify for travis, and reduce our builds by 3

DivineDominion:
Settings like `FRAMEWORK_SEARCH_PATHS[sdk=watchos*]` wouldn't work for the simulator, but we don't use that, so out with it :) Thanks for pointing out the redundancy, @mjarvis

PR #390, opened 2019-01-28
Support swift 5.0.

kawoou:
✨Supported Swift 5.0 ✨


ReSwift/CoreTypes/Store.swift
```diff
@@ -134,9 +134,15 @@ open class Store<State: StateType>: StoreType {
     }
 
     open func unsubscribe(_ subscriber: AnyStoreSubscriber) {
+        #if compiler(<5.0)
```
mjarvis: In order to correctly support differing compilers, this should be `#if compiler(>=5.0)`, and the if/else contents should be reversed

mjarvis:
@kawoou Many thanks for your contribution! I'm very excited for Swift 5.0, even though its going to finally force me to update my computer to 10.14...

ReSwift/CoreTypes/Store.swift
mjarvis: Actually, to go a bit further than this, we'll need this to use an `#if swift(...)` check, since we still support Swift 3.2, which does not have the `compiler` directive. I think I'm going to open discussion elsewhere for dropping Swift 3.2.

ReSwift/CoreTypes/Store.swift
```diff
@@ -134,9 +134,15 @@ open class Store<State: StateType>: StoreType {
     }
 
     open func unsubscribe(_ subscriber: AnyStoreSubscriber) {
+        #if swift(<5.0)
```
mjarvis: Until https://github.com/apple/swift/commit/e66095b10a8caa8ab568efc36223835120a1f7fc , the compiler directives did not support the `<` operator. This change is only available in Swift 5.0, so while this works in 5.0, it fails in anything below, as seen by the Travis CI jobs.

In order for this to work you'll need to switch it to `#if swift(>=5.0)`, and swap the contents of the if/else.

kawoou: Thank you for telling me.
I'll fix it :)

mjarvis:
@kawoou Thanks for the updates. I'm going to hold on merging this until Travis CI supports Xcode 10.2

aaronsky:
Just a heads up that folks depending on this project via Carthage need this change in order to build. Hopefully Travis puts out beta support soon!

nh7a:
**Supported Swift Versions: Swift 3.2, 4.0** in README.md needs to be modified as well since Swift 3 mode will be deprecated by up coming Xcode.

mjarvis:
Closing in favour of https://github.com/ReSwift/ReSwift/pull/397 which resolves xcode's warning about base localization, as well as conditional conformance to the new `Hashable` setup.

PR #391, opened 2019-02-01
Deprecate `ActionCreator` and `AsyncActionCreator` (#391)

mjarvis:
`ActionCreator` and `AsyncActionCreator` are deprecated in favor of https://github.com/ReSwift/ReSwift-Thunk

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@
     - Removes the ability to directly set `state` by making it `private(set)`. This prevents users from bypassing reducers and middleware. All mutation of the state must occur through the normal `Action` & `Reducer` methods.
     - This deprecates the usage of `ReSwift-Recorder`. Changes may be made to that library in the future in order to support this change.
 
+- Deprecate `ActionCreator` and `AsyncActionCreator` (#391) - @mjarvis
+
+    - These are deprecated in favor of https://github.com/ReSwift/ReSwift-Thunk
+
 **Other**:
 - Add Subscription `skip(when:)` and `only(when:)` (#242) - @mjarvis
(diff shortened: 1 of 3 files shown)


brokenseal:
I have been using ReSwift for a few years now for a project I picked up again just yesterday.
This change comes as a real surprise, I would encourage you guys to rethink this one since having external plugins is nice, but you should never force users to use them. This warning message in my IDE has become very annoying... :)

danielmartinprieto:
Hi, @brokenseal. Note that you don’t need ReSwift-Thunk, it’s a utility and that’s why it’s been put in an external module. You can perform any kind of side effect just using a little middleware.

The original design of action creators and async action creators wasn’t correct and people had a lot of questions and misunderstandings (see the issues, there are tons with a lot of discussions about it) and that’s why they’ve been deprecated.

As with every piece of open source software, you have several options so you don’t see those warnings: you can change your code to be up to date with ReSwift (preferable), you can use an older version that matches your code (not recommended, but it could work for until you have time to upgrade properly) and you can even fork the library and adapt it to a more concrete use case.

Don’t hesitate to post an issue if you need help migrating away from the old action creators, we’ll be happy to help!

brokenseal:
Hi,
thanks a lot for the eloquent description, much appreciated. I understand this is an open source project and that there are several way to circumvent any issue.

So, since it's flagged as deprecated, it will be removed in a future version?

mjarvis:
To be explicit, they have already been removed in latest master, and that change will be included in the next major release (5.0) -- See https://github.com/ReSwift/ReSwift/blob/master/CHANGELOG.md for details.

brokenseal:
Thanks for the clarification and thanks for your effort put into this open source library :)

PR #397, opened 2019-03-15
Resolve Xcode 10.2 warnings with Swift 4.2.2 and 5.0

### LINKED ISSUE #396: Swift 4.2.2 & Swift 5.0 Support
ReSwift should be updated to conditionally work with the Xcode 10.2 beta, which uses Swift 5.0 or Swift 4.2.2 as a compatibility mode to 4.2 code.

mjarvis:
Resolves #396

For compatibility reasons, `swift_version` has been left at 4.2. This works in both Xcode 10.1, as well as in Xcode 10.2 (which uses Swift 4.2.2), and is thus also cross-compatible with Swift 5.0.
I see no reason for us to update the project version to swift 5.0 as long as the compatibility mode exists, as Carthage users can mix 4.2.2 libraries with 5.0 code, and for CocoaPods it will use their specified swift version, not ours.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![Build Status](https://img.shields.io/travis/ReSwift/ReSwift/master.svg?style=flat-square)](https://travis-ci.org/ReSwift/ReSwift) [![Code coverage status](https://img.shields.io/codecov/c/github/ReSwift/ReSwift.svg?style=flat-square)](http://codecov.io/github/ReSwift/ReSwift) [![CocoaPods Compatible](https://img.shields.io/cocoapods/v/ReSwift.svg?style=flat-square)](https://cocoapods.org/pods/ReSwift) [![Platform support](https://img.shields.io/badge/platform-ios%20%7C%20osx%20%7C%20tvos%20%7C%20watchos-lightgrey.svg?style=flat-square)](https://github.com/ReSwift/ReSwift/blob/master/LICENSE.md) [![License MIT](https://img.shields.io/badge/license-MIT-blue.svg?style=flat-square)](https://github.com/ReSwift/ReSwift/blob/master/LICENSE.md) [![Reviewed by Hound](https://img.shields.io/badge/Reviewed_by-Hound-8E64B0.svg?style=flat-square)](https://houndci.com)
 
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -2,7 +2,7 @@
 
 [![Build Status](https://img.shields.io/travis/ReSwift/ReSwift/master.svg?style=flat-square)](https://travis-ci.org/ReSwift/ReSwift) [![Code coverage status](https://img.shields.io/codecov/c/github/ReSwift/ReSwift.svg?style=flat-square)](http://codecov.io/github/ReSwift/ReSwift) [![CocoaPods Compatible](https://img.shields.io/cocoapods/v/ReSwift.svg?style=flat-square)](https://cocoapods.org/pods/ReSwift) [![Platform support](https://img.shields.io/badge/platform-ios%20%7C%20osx%20%7C%20tvos%20%7C%20watchos-lightgrey.svg?style=flat-square)](https://github.com/ReSwift/ReSwift/blob/master/LICENSE.md) [![License MIT](https://img.shields.io/badge/license-MIT-blue.svg?style=flat-square)](https://github.com/ReSwift/ReSwift/blob/master/LICENSE.md) [![Reviewed by Hound](https://img.shields.io/badge/Reviewed_by-Hound-8E64B0.svg?style=flat-square)](https://houndci.com)
 
-**Supported Swift Versions:** Swift 3.2, 4.0
+**Supported Swift Versions:** Swift 3.2, 4.0, 4.2, 5.0
```
nh7a:
```suggestion
**Supported Swift Versions:** Swift 4.0, 4.2, 5.0
```
Since Xcode 10.2 doesn't support Swift 3.2.

mjarvis: It is still possible to use ReSwift with older versions of Xcode via CocoaPods, or manual integration.
I will look into dropping Swift 3.2 for our next major release but we should maintain backwards compatibility for the moment.

danielmartinprieto:
I'm not using Xcode 10.2 yet, but I agree with your comment in this PR. If it's compatible with Swift 5.0, just fixing the warnings is enough to move on for now, although it's always nice to use the last version of the language (so changes in the future are easier), but we can do that later. Apart from that, I've reviewed the changes and they lgtm.

Issue #394, opened 2019-03-15
Possible `automaticallySkipsRepeats` regression on `master`


peter-tomaselli:
According to [this comment](https://github.com/ReSwift/ReSwift/issues/303#issuecomment-461656238), there is some problem with `automaticallySkipsRepeats` not actually functioning correctly for simple `Equatable` states on the `master` branch, as compared to the `4.0.1` release.

I tried to reproduce this problem myself with some simple unit tests — in the style of `AutomaticallySkipRepeatsTests.swift` — but was unable to do so. I’m assuming that’s because this is either not a problem anymore, was perhaps never really a problem, or most likely, more detail is needed to reproduce the original bug report.

I was encouraged to file an issue here anyway 😄 because I was told that the possibility of this bug was blocking the release of [`4.1.0`](https://github.com/ReSwift/ReSwift/milestone/3), which my team would love to see released. (Specifically, we’d really like to have a Swift-4.2-compatible release tag to pin our dependencies to, instead of having to use `master`).

If there’s anything else I can do to help track down this bug, or to facilitate a new release in general, please let me know. Thank you!!

mjarvis:
@sjmueller This ticket is in regards to your comments.

To note: @DivineDominion made an incorrect statement regarding the default value for `automaticallySkipsRepeats`. The commit that the linked to correctly shows the default value is `true`. And can also be confirmed by looking at current master: https://github.com/ReSwift/ReSwift/blob/master/ReSwift/CoreTypes/Store.swift#L60

I am also unable to reproduce this issue. I attempted to reproduce by adding the following test:

```swift
    func testSkipsStateUpdatesForEquatableSubStateByDefault() {
        let reducer = TestValueStringReducer()
        let state = TestStringAppState()
        let store = Store(reducer: reducer.handleAction, state: state, middleware: [])
        let subscriber = TestFilteredSubscriber<String>()

        store.subscribe(subscriber) {
            $0.select { $0.testValue }
        }

        XCTAssertEqual(subscriber.receivedValue, "Initial")

        store.dispatch(SetValueStringAction("Initial"))

        XCTAssertEqual(subscriber.receivedValue, "Initial")
        XCTAssertEqual(subscriber.newStateCallCount, 1)
    }
```

This test passes on current master. If the skip was not occurring the last line would fail, which can be confirmed by setting `automaticallySkipsRepeats: false` on the store in the test.

mjarvis:
[...]

PR #399, opened 2019-03-21
fix automaticallySkipsRepeats when store has a state that is not equatable, but the selected substate is equatable (possible regression in 4.1.0)

djtech42:
It seems like it should look at if the substate from the transform is `Equatable` instead of both the state and substate for `subscribe`. 

When updating to 4.1.0, some substates stopped using `skipRepeats()` even though they were `Equatable` because the store's main state is not `Equatable`. The configuration `automaticallySkipsRepeats` can still be used to turn off automatic skipping for this case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 - Improve subscription & state update performance (#325) - @mjarvis
 - Enable build settings "Allow app extension API only" (#328) - @oradyvan
 - Open `Subscription<State>` to allow external extensions (#383) - @mjarvis
+- Fix 4.1.0 regression with `automaticallySkipsRepeats` when selecting Equatable state in Non-Equatable root state (#399) - @djtech42
 
 # 4.0.1
 
--- a/ReSwift/CoreTypes/Store.swift
+++ b/ReSwift/CoreTypes/Store.swift
@@ -201,16 +201,7 @@ open class Store<State: StateType>: StoreType {
 
 // MARK: Skip Repeats for Equatable States
 
-extension Store where State: Equatable {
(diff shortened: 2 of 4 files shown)


mjarvis:
Ah! This I believe is related to https://github.com/ReSwift/ReSwift/issues/394
I'm going to see if I can reproduce this myself as well to confirm

djtech42:
I added some unit tests for this case now

mjarvis:
Heres a more simple unit test for this case that fits in with our other subscription tests.
Please replace your tests changes with this:
```swift
    func testSkipsUpdatesForEquatableSubStateByDefault() {
        let reducer = TestNonEquatableReducer()
        let state = TestNonEquatable()
        let store = Store(reducer: reducer.handleAction, state: state)
        let subscriber = TestFilteredSubscriber<String>()

        store.subscribe(subscriber) {
            $0.select { $0.testValue.testValue }
        }

        XCTAssertEqual(subscriber.receivedValue, "Initial")

        store.dispatch(SetValueStringAction("Initial"))

        XCTAssertEqual(subscriber.receivedValue, "Initial")
        XCTAssertEqual(subscriber.newStateCallCount, 1)
    }
```

mjarvis:
@djtech42 Hey just wanted to let you know, this is great. I'll be approving & merging, and releasing 4.1.1 soon. I'm just working on figuring out in which change this regression occurred, now that we have a reproducible test case.

mjarvis:
Looks like the regression was caused in bdb648 which was merged in #262

DivineDominion:
Great work!

Issue #318, opened 2018-01-12
How to subscribe to multiple states separately in one class


gkmrakesh:
ReSwift Version: 4.0.0

**Problem: I have to listen to multiple sub-states separately in one class**

Currently, i am doing it like:

```
class Sample: StoreSubscriber {
   init() {
         // MARK: Subscriber of state 1, 2
         appStore.subscribe(self) { state in
            state.select { state in (state.state1, state.state2) }
        }
   }

   // MARK: Listener for state 1, 2
   func newState(state: (state1: State1, state2: State2) ) {
       // process
   }

}
```

**How can i listen to each sub-state separately, so that i can act on only sub-state which get changed??**

SomethingLike:  

```
class Sample: StoreSubscriber {
   init() {
         // MARK: Subscriber
         appStore.subscribe(self) { state in
            state.select { state in state.state1 }
        }

         appStore.subscribe(self) { state in
            state.select { state in state.state2 }
        }
   }

   // MARK: Listener for state1
   func newState(state: State1 ) {
       // process
   }

   // MARK: Listener for state2
   func newState(state: State2 ) {
       // process
   }

}
```

DivineDominion:
You can't; you'd need to write 2 subscribers and 1 combined state change handler as separate objects.

That being said, I wonder if your state is partitioned well if you run into this. Can you give a more concrete example?

gkmrakesh:
@mbroski , @DivineDominion ,

Thanks for your quick reply.

@mjarvis ,

**Awesome, after months of struggle I can able to separate subscribers.**

But, issue remains same,

When my sub-state changes, all subscribers (state1Subscriber, state2Subscriber in below code) getting notified.

This brings me back to initial state. 

How can I listen to only sub-state which get changed?? 

For example in below code, if "Counter" state changed then only state1Subscriber will get notified.

```
import UIKit
import ReSwift

public class BlockSubscriber<S>: StoreSubscriber {
    public typealias StoreSubscriberStateType = S
    private let block: (S) -> Void
    
    public init(block: @escaping (S) -> Void) {
        self.block = block
    }
    
    public func newState(state: S) {
        self.block(state)
    }
}

class ViewController: UIViewController {
    
    @IBOutlet weak var counterLabel: UILabel!
    @IBOutlet weak var counterLabel2: UILabel!
    
    private lazy var state1Subscriber: BlockSubscriber<Counter> = BlockSubscriber(block: { [unowned self] state1 in
        print(state1)
    })
[...]

PR #401, opened 2019-03-29
Update readme with info about how to create a combined subscription

danielmartinprieto:
As per @DivineDominion's request [here](https://github.com/ReSwift/ReSwift/issues/318/), I've added a subsection with a brief explanation of how to create a subscription combining several substates easily.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,7 +135,27 @@ Button taps result in dispatched actions that will be handled by the store and i
 
 This is a very basic example that only shows a subset of ReSwift's features, read the Getting Started Guide to see how you can build entire apps with this architecture. For a complete implementation of this example see the [CounterExample](https://github.com/ReSwift/CounterExample) project.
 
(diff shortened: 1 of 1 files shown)


DivineDominion:
Looking at the README now, I think a couple more lines of code would be useful: what's the `AppState` like, and how does the `MySubState` assemble a view into it?

Then it might be a better fit for the guides instead of the front page README. What do you think?

DivineDominion:
@danielmartinprieto What do you say about the suggestions? Do you think it makes sense to move this from the README to the guides/help files, too?

danielmartinprieto:
Sorry, I didn't have time to do this. Yeah, it's totally fine, we can later link some FAQs from the README to those guides.

danielmartinprieto:
You can move it already if you want, and we'll improve it later. Better having something (even if it's not complete) that nothing at all.

DivineDominion:
You're right, doesn't do any harm for now.

PR #403, opened 2019-04-01
Update SwiftPM manifest file

Dschee:
This adds support for SwiftPM manifest based dependency managers. Specifically this adds support for installing via [Accio](https://github.com/JamitLabs/Accio) but will probably also work with SwiftPM once it's integrated into Xcode.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,22 +28,28 @@ matrix:
       env: SCHEME="watchOS"  SDK="watchsimulator"     DESTINATION="OS=4.0,name=Apple Watch - 42mm" SWIFT_VERSION="4.2" ACTION="build"
     - osx_image: xcode10.1
       env: SCHEME="tvOS"     SDK="appletvsimulator"   DESTINATION="OS=12.0,name=Apple TV 4K"       SWIFT_VERSION="4.2" ACTION="test"
+    - osx_image: xcode10.2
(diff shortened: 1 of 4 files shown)


mjarvis:
@Dschee Many thanks for your contribution! Is there a way that we can have a CI job to lint / build via swift PM to ensure that this is setup correctly?

Dschee:
Hey @mjarvis, of course there are ways to configure SwiftPM support and Accio support are checked on the CI. I'm not sure how to do it in Travis CI though, but the basic ideas are:

**For testing SwiftPM support:**
Run `swift test` from the projects root directory of a macOS machine having at least Xcode 10.1 installed.

**For testing Linux support:**
The same as for SwiftPM support, but on a Ubuntu machine after installing the Swift Ubuntu toolset.

Accio support is automatically given if you support SwiftPM _and_ Carthage. Actually, currently there is no command to test the support without creating a Package.swift file that includes your repo like I've already setup in the [Demo](https://github.com/JamitLabs/Accio/tree/work/1000-frameworks/Demo) project at Accio. But I just created https://github.com/JamitLabs/Accio/issues/16 to make sure we add such a feature over time.

Dschee:
I just looked into some Travis CI config files and found that PromiseKit is testing SwiftPM support by running `swift build` (which basically tests installability and skips the tests) [here](https://github.com/mxcl/PromiseKit/blob/master/.travis.yml#L176). You can also find a way on how you could add Linux tests there, too.

I hope that helps.

mjarvis:
Many thanks I'll look those over and update our CI config.

PR #405, opened 2019-04-06
Update Swift Package Manager to v4.2

jayhickey:
Swift 5 no longer supports the Swift 3 Package.swift tools-version, which means Xcode 10.2 doesn't either. This updates Package.swift to the latest, version 4.2.


jayhickey:
Oh, looks like this is already done in #403

PR #411, opened 2019-06-04
add swift package manager support for 5.0

hoemoon:
I tried to use reswift from xcode11 with swift package manager .
I noticed that there is no swift package metafile for 5.0.


DivineDominion:
@hoemoon Cool! Could you fix the SwiftLint issues and maybe update the outdated instructions in the docs regarding SwiftPM while you are working on the topic?

hoemoon:
I updated readme for SPM support and fixed SwiftLint issue.

Issue #237, opened 2017-04-22
ReSwift crashes on device (but not simulator)


snydersaurus:
Help!  I have been chugging along using ReSwift (on 3.0, Swift3, XCode 8.3) and now for some reason upon initialization ReSwift is crashing (when I run it on my 10.3.1/iPhone7+ thru XCode). I hate these esoteric errors and am hoping beyond hope that someone has experienced anything remotely similar. 

AppDelegate has the usual declaration
```
let store = Store<State>(reducer: AppReducer(), state: nil)
```

Then my first ViewController subscribes:

```
override func viewWillAppear(_ animated: Bool) {
        super.viewWillAppear(animated)
        
        store.subscribe(self) { state in
            state.authenticationState
        }
 }
```

ReSwift goes through the initialization process until it gets to my AppReducer, and BAM !

<img width="971" alt="screen shot 2017-04-21 at 10 30 24 pm" src="https://cloud.githubusercontent.com/assets/2645768/25300680/f0fa6486-26e2-11e7-8c52-86394112a037.png">

Thanks everyone,

--Bill

snydersaurus:
I am noticing this device crash happens in Debug, but not Release. Are there optimizations or other build/compiler settings that would cause problems with ReSwift?

snydersaurus:
Well for posterity -- I am further noticing that in Debug with No Optimization I crash. Flipped the Swift/LLVM optimization levels to -O and no crash.

buscarini:
I'm also seeing this. I think it's related to this compiler bug:

https://bugs.swift.org/browse/SR-4432

Seems to happen when assigning or passing big structs as arguments. This is a big problem for ReSwift based applications, because with this architecture it's unavoidable having a big state. We have this issue in one of our apps (but not on others). This app is the biggest where we have used this architecture, so probably that's why we are hitting it.

Please reopen this issue. I think it's important that people that use ReSwift are aware of this problem. For now the bug is still open in Apple's Jira; let's hope it gets fixed for Swift 4 release.

snydersaurus:
Reopened. 
Also still an issue for me on the latest stable tooling (Swift 3.1, XCode 8.3.3). 

Note @buscarini that my initial AuthenticationState struct is small, it does have a few simple child structs. Nesting is not deeper than that though.

buscarini:
[...]

PR #414, opened 2019-07-10
update project for Swift 5

DivineDominion:


mjarvis:
I recommend we leave the pbxproj swift version set to 4.2. See description in #397 regarding Carthage compatibility.

DivineDominion:
Gotcha. Since #237 isn't affected by this anyway, it seems, there's no reason to push this. Also no upside of leaving this PR dangling because the changes are trivial.

Issue #289, opened 2017-09-21
Support subscribing with keypaths in Swift 4


kdawgwilk:
It would be nice to be able to subscribe to more specific items in the state using keypaths added in Swift 4.  Example of how they were adopted in The SDKs here: https://developer.apple.com/library/content/documentation/Swift/Conceptual/BuildingCocoaApps/AdoptingCocoaDesignPatterns.html#//apple_ref/doc/uid/TP40014216-CH7-ID12

danielmartinprieto:
Hi, @kdawgwilk. Could you explain a little more how the use of keypaths would improve the library? Thanks!

kdawgwilk:
Yeah so here is a brief example:

## Goal

Be able to subscribe with finer grain using KeyPaths like this:
```swift
store.subscribe(self, to: \AppState.someSubState.property)
```

## Implementation Example

Added `Store.swift` method:
```swift
open func subscribe<V, S: StoreSubscriber>(
    _ subscriber: S, to keyPath: KeyPath<State, V>
) where S.StoreSubscriberStateType == V {
    // If the same subscriber is already registered with the store, replace the existing
    // subscription with the new one.
    if let index = subscriptions.index(where: { $0.subscriber === subscriber }) {
        subscriptions.remove(at: index)
    }

    let value = state[keyPath: keyPath]
    print("Value being observed: \(value)")

    // Details left out
}
```

kdawgwilk:
The complete implementation might look like this?
``swift
open func subscribe<V: StateType, S: StoreSubscriber>(
_ subscriber: S, to keyPath: KeyPath<State, V>
) where S.StoreSubscriberStateType == V {
    // If the same subscriber is already registered with the store, replace the existing
    // subscription with the new one.
    if let index = subscriptions.index(where: { $0.subscriber === subscriber }) {
        subscriptions.remove(at: index)
    }

    // Create a subscription for the new subscriber.
    let originalSubscription = Subscription<State>()
    let transformedSubscription = Subscription<V>()

    let subscriptionBox = SubscriptionBox(
        originalSubscription: originalSubscription,
        transformedSubscription: transformedSubscription,
        subscriber: subscriber
    )

    subscriptions.append(subscriptionBox)

    if let state = self.state {
        originalSubscription.newValues(oldState: nil, newState: state)
    }
}
```

svdo:
FYI: I also support this in my library https://github.com/svdo/ReRxSwift.

[...]

PR #415, opened 2019-07-11
add key paths to subscription select

djtech42:
Example: `store.subscribe(self) { $0.select(\.name) }`

### DIFF
--- a/ReSwiftTests/AutomaticallySkipRepeatsTests.swift
+++ b/ReSwiftTests/AutomaticallySkipRepeatsTests.swift
@@ -28,19 +28,19 @@ class AutomaticallySkipRepeatsTests: XCTestCase {
     }
 
     func testInitialSubscription() {
-        store.subscribe(self) { $0.select { $0.name } }
+        store.subscribe(self) { $0.select(\.name) }
         XCTAssertEqual(self.subscriptionUpdates, 1)
     }
 
     func testDispatchUnrelatedActionWithExplicitSkipRepeats() {
-        store.subscribe(self) { $0.select { $0.name }.skipRepeats() }
+        store.subscribe(self) { $0.select(\.name).skipRepeats() }
         XCTAssertEqual(self.subscriptionUpdates, 1)
         store.dispatch(ChangeAge(newAge: 30))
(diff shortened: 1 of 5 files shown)


DivineDominion:
@djtech42 Could you resolve the linter conflict and add an entry for your PR to the Changelog? 

I like the API, and I cannot find anything wrong with this approach, so I'd merge this in. I do recall a discussion related to key paths, but cannot find anything here. Does this trigger your memory, or do you have objections @mjarvis ?

danielmartinprieto:
Is it ok that we change tests to use keypaths instead of adding a few new ones?

djtech42:
It looks like key paths were discussed here before: https://github.com/ReSwift/ReSwift/issues/289

ReSwiftTests/AutomaticallySkipRepeatsTests.swift
```diff
@@ -28,19 +28,19 @@ class AutomaticallySkipRepeatsTests: XCTestCase {
     }
 
     func testInitialSubscription() {
-        store.subscribe(self) { $0.select { $0.name } }
+        store.subscribe(self) { $0.select(\.name) }
```
DivineDominion: can you duplicate these tests instead of replacing the old ones? Then we have a way to ensure that both implementations do the same and catch regressions

ReSwiftTests/StoreSubscriberTests.swift
```diff
@@ -116,8 +141,9 @@ class StoreSubscriberTests: XCTestCase {
         let subscriber = TestFilteredSubscriber<TestCustomAppState.TestCustomSubstate>()
 
         store.subscribe(subscriber) {
-            $0.select { $0.substate }
-                .skipRepeats { $0.value == $1.value }
+            $0
```
DivineDominion: same here: rather introduce a `_RegularSubstateSelection_` and `_WithKeyPaths_` in the test names to disambiguate the duplicates than remove the old ones

DivineDominion:
had to re-run travis, but now it passes. Thanks, @djtech42 !

PR #418, opened 2019-08-04
remove test variants for Swift 4.1 and lower

### LINKED ISSUE #419: Drop support for Swift 3, 4.1
I think we can drop support for these years-old versions. Do you have objections?

DivineDominion:
I think we should be dropping support for Swift before 4.2 eventually. I removed conditional compiling instructions for Swift 4.1 and earlier. See #419 for the discussion

Closes #419.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,10 +16,6 @@ env:
 
 matrix:
   include:
-    - osx_image: xcode9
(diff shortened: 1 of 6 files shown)


mjarvis:
I'll take a look at this tomorrow, and make sure docs & things are updated appropriately for the dropped versions.

DivineDominion:
@mjarvis Thanks for supplying the missing pieces! Do you still want to do anything else, should I fix something that's missing, or are we good to go?

mjarvis:
@DivineDominion I think we're good to go.

PR #425, opened 2019-09-03
Run pod linter in Travis

DivineDominion:
I attached the new action to the configuration that tests SwiftPM, and skip it during all the others. We could just as well put it into the latest iPhone's configuration; I'm not sure if there's any difference.


DivineDominion:
Pushed the change based off of a wrong branch at first, please ignore the bots' comments :)

PR #427, opened 2019-09-26
Add capability to mutate middleware after store creation.

### LINKED ISSUE #426: Mutable Middleware
In some cases, it can be useful to add or remove middleware after the Store is initialized, such as adding logging middleware while attempting to debug an issue, or removing middleware in order to playback raw actions without side-effects.

At first glance, it appears possible to make a `setMiddleware(...)` function which re-write the dispatch function accordingly.

It might be desirable to extend this somehow to have middleware be a property of the store, but this may have unforeseen consequences or requirements. I will attempt to implement this, falling back to more restrictive `setMiddleware()` if issues arise.

PR to come.

mjarvis:
Resolves #426

In some cases, it can be useful to add or remove middleware after the Store is initialized, such as adding logging middleware while attempting to debug an issue, or removing middleware in order to playback raw actions without side-effects.

This exposes `public var middleware: [Middleware<State>]` to allow standard array mutation of the middleware at anytime in the Store's lifecycle.

We could also remove the implicit unwrapping on `var dispatchFunction: DispatchFunction!`, however I've chosen to leave it in in order to prevent any unexpected breaking changes if users were accessing or setting this variable, since it is `public`.

`didSet` is not called on variables if they are set during the init process of a class, and thus we need to manually create the dispatch function for the first access. This is solved by making `dispatchFunction` `lazy`, so that we do not need to assign it in the initializer.

This commit is a purely additive change, and should have zero effect on existing functionality.

TODO:
- [x] Add tests to ensure mutation of middleware is applied correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# Upcoming
+
+**API Changes:**
+- Add capability to mutate `Middleware` after store creation. (#) - @mjarvis
+
 # 5.0.0
 
 *Released: 2019-06-30* 
--- a/ReSwift/CoreTypes/Store.swift
+++ b/ReSwift/CoreTypes/Store.swift
@@ -76,7 +76,7 @@ open class Store<State: StateType>: StoreType {
         }
     }
 
-    private func createDispatchFunction() -> DispatchFunction {
+    private func createDispatchFunction() -> DispatchFunction! {
         // Wrap the dispatch function with all middlewares
         return middleware
             .reversed()
--- a/ReSwift/CoreTypes/Store.swift
+++ b/ReSwift/CoreTypes/Store.swift
@@ -29,7 +29,7 @@ open class Store<State: StateType>: StoreType {
         }
     }
 
-    public var dispatchFunction: DispatchFunction!
+    public lazy var dispatchFunction: DispatchFunction! = createDispatchFunction()
 
     private var reducer: Reducer<State>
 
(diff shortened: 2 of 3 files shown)


DivineDominion:
Cool, sounds good. But also means I have to update #420 with the changes. Hope nothing breaks there. (Shameless plug: feedback on that is highly welcome :))

DivineDominion:
Looks good so far. How do you remove Middleware in practice, like remove something from the middle of the sequence? I was thinking about using names, because you can only `remove` from an Array using the index.

mjarvis:
@DivineDominion I thought about this for a while... I think the current form of Middleware doesn't allow for easy removals. We'd have to change the definition, and I'd really want to be sure of the value before making a bigger breaking change like that.
The biggest use cases I see for this are adding middleware after initialization, removal of all middleware to allow for pure replay of actions, and re-adding all middleware to reset after replay.

danielmartinprieto:
Although I’m not opposed to the change, I don’t think updating the global store like that is the best/safest approach to handle the examples you mentioned.

I never encountered a use case like this in an app (neither in Swift or JavaScript), but if I had to reply some actions, I would probably want the side effects associated to the actions applied, and otherwise I would wrap the original action in one of type “reply” (for instance) so the intention is clear.

The only reason I would use something like that is to avoid the runtime overhead of a middleware that I only use in a sub part of my app, but as I said before, I’m not sure that fits well with the “global store” paradigm. And definitely Redux favors immutability in general, so that would be also an counter argument.

Also, these are my thoughts but maybe I’m missing something fundamental.

PR #442, opened 2020-05-18
Update Store's state setter to be internal, to allow state to be mutated directly from unit tests

johntmcintosh:
There was a change awhile back to remove the public setter from the `Store`'s `state` property to encourage better state access patters. 

Before that change was made, we found that it was very helpful for us to be able to directly mutate a value in the store's state from our unit tests as a precursor to testing the effects of a particular reducer. 

Would you all be open to slightly reducing the restriction here so that the store's state setter was `internal` rather than `private`? This would allow us to `@testable import ReSwift` in our unit testes and modify the state directly, but still have the tighter restrictions on production code.

### DIFF
--- a/ReSwift/CoreTypes/Store.swift
+++ b/ReSwift/CoreTypes/Store.swift
@@ -17,7 +17,7 @@ open class Store<State: StateType>: StoreType {
 
     typealias SubscriptionType = SubscriptionBox<State>
 
-    private(set) public var state: State! {
+    internal(set) public var state: State! {
         didSet {
             subscriptions.forEach {
                 if $0.subscriber == nil {


dani-mp:
Hi, could you elaborate on _testing the effects of a particular reducer_? A reducer is just a pure function so it should be quite easy to test on its own.

johntmcintosh:
Hi @dani-mp , you're right, I misspoke a bit in my initial comment. The test cases that we are writing are really more of integration tests than pure unit test around reduce functions. We will often do something this this in test cases:

- Construct an initial state
- Dispatch an action
- Assert that the updated state is correct
- Assert that various other things happened because of middleware (for example, we use middleware to trigger analytic logs for certain actions, or to dispatch other actions for related things that need to happen)

We typically end up with test cases that look something like this:

```
class MyTests: XCTestCase { 
    var store: TestStore!
    var analyticsService: MockAnalyticsService!

    override func setUp() { 
        super.setUp()
        analyticsService = MockAnalyticsService()
        store = makeTestStore(analyticsService)
    }

    func testMyAction() { 
        store.state.property1 = "initialValue"
        store.state.property2 = "anotherInitialValue"

        store.dispatch(MyAction())

        XCTAssertEqual(store.state.property1, "newValue")
        XCTAssertEqual(store.state.property2, "anotherNewValue")
        XCTAssertTrue(analyticsService.lastEvent is MyAnalyticEvent)
    }
}
```

It would technically be possible for us to test each of these independently, but we have found that it works quite well for us to test these in one place using a structure like ^. 

I think that we could also create a new store instance with our initial test state rather than mutating the state within the store, but it's less convenient since we couldn't rely on setting up the store in our `setUp` override (and we're trying to keep the tests as convenient as possible 😀).

dani-mp:
@johntmcintosh thanks for sharing an example.

I don't think that what you're doing there is a good idea in order to have reliable integration tests. In your test, when you do `store.dispatch(MyAction())` you're assuming that your state looks like the one you manually set above, but in your real application that may be different because to reach that point you had to dispatch other actions that affect the store in different ways, and you're not taking those into account here. That's why for a good integration test you should test your store using its public API, the one your application is going to use, including the rest of the reducers and middleware. You can always create helper functions that put (using public APIs) your store in a concrete state to make your tests less verbose.

And as we commented before, you can always test each part independently with unit tests.

mjarvis:
With the way `XCTest` works, `setUp()` is called before every test function is executed, so you could resolve your needs here by initializing the store in the test function itself, with the state pre-set in the form that you require, rather than creating the store and then mutating the state.

johntmcintosh:
Hi guys, thanks for the feedback. You have valid points, but without getting too far into the weeds to debate test philosophies I think the conversation is missing the broader point that I mean to be making.

With the state having a `private` setter from the `store`, we are limited in terms of what we can do from our tests. I fully support keeping the setter private for actual usage. Opening it up to be `internal` would be no different than `private` for regular use, but would allow the setter to be accessed from test targets, giving people more flexibility to write tests in a way that works best for their team. 

@mjarvis, I thought about your suggestion, and it would work for us for many cases, although it would add additional coding overhead to every test case. (Not functional overhead, just visual noise.) However, we have a fairly large app (several thousand test cases) and there are a number of more complex tests where it has been effective for us to combine a couple of things into a single test where we: construct a state, dispatch an action, make a couple of assertions, manually tweak something in the state, dispatch another action, make some more assertions. One can argue that we shouldn't be testing that way, but for the nature of what we're testing in those scenarios, it has been a concise and practical way to validate what we needed.

Do you have particular concerns with allowing test targets write access to the store?

mjarvis:
My concerns around allowing test targets write access to the state have to do with avoid the regular steps taken in the `action -> middleware -> reducers -> middleware -> subscribers` path.

Allowing mutation of state breaks the regular path, as there is a `didSet` on state which calls subscribers.

This opens up potential for unintended consequences if a user has a test store with middleware and /or reducers and sets state without considering it skipping those devices.

At the moment the consequences are low, but opening up `state` to setting in test targets adds another "public" (via `@testable import`) interface that we have to consider when avoiding breaking changes. (IE imagine we move the subscriber calls out of the didSet, or some similar change)

dani-mp:
Adding to what @mjarvis said, and insisting on my previous comment, those test would not be telling you the whole story and could lead to misconceptions regarding how your store works in your app vs. in your tests.

I understand you are aware of this and are ok with it, but I don’t think it’s a good idea in general because it’s not how ReSwift is supposed to be used.

johntmcintosh:
Thanks, I appreciate the additional context. I'll go ahead and close this out.

PR #446, opened 2020-08-18
Use GCD instead of deprecated OS level atomic API

basememara:
Adding concurrency handling was a good call in #341. However, opting for GCD is always better because speed is not always the factor to measure performance. Resource management such as memory and CPU is fine tuned with GCD. For example, `OS` level API's can use 1000+% CPU while GCD will make sure it doesn't go too far over 100%. In the GCD implementation provided here, concurrent reads are happening for fast performance while reading, but a barrier is placed around the writes so no reads or writes can occur during a write. Also `OSAtomicTestAndSet` [seems to be deprecated](https://stackoverflow.com/questions/44685236/swift-3-atomic-boolean).

I added a link in the `Synchronized` utility for reference in case anyone in the future has questions about it or plan to change it, but I can remove it if you like. I've also added unit tests around the synchronization itself, but will a low number to not hold up the unit tests.


mjarvis:
@basememara Could you provide an example run of performance tests before & after this change?

basememara:
> Is it possible to keep the previous API?
> 
> ```swift
> isDispatching.value = false
> // instead of
> isDispatching.value { $0 = false }
> ```

As far as I know it is not possible because it needs to be in a closure so it gets passed into the queue.

basememara:
@mjarvis good call, let me think of a way to hammer ReSwift in a unit test

mjarvis:
@basememara There is a very basic dispatch performance test included in project in `ReSwiftTests/PerformanceTests.swift` that may be sufficient for at least the simple time-based comparison.

ReSwift/Utils/Synchronized.swift
```diff
@@ -0,0 +1,26 @@
...
+
+/// An object that manages the execution of tasks atomically.
+struct Synchronized<Value> {
+    private let mutex = DispatchQueue(label: "reswift.github.io.ReSwift.Utils.Synchronized", attributes: .concurrent)
+    private var _value: Value
+    init(_ value: Value) {
+        self._value = value
+    }
+    /// Returns or modify the thread-safe value.
+    var value: Value { mutex.sync { _value } }
```
mjarvis, line 21: For the request by @dani-mp maybe this will work. We lose the ability to throw in the setter but we have no need for that in this use case.
```suggestion
    var value: Value {
        get { mutex.sync { _value } }
        set { mutex.sync(flags: .barrier) { _value = newValue }
    }
```

basememara: Unfortunately, there was something it didn't like about that. The unit test failed for the concurrency with `Thread 9: EXC_BAD_ACCESS (code=EXC_I386_GPFLT)`. I think the `+=` expression tricks it and no longer part of the closure in the setter. This is what stopped me from creating it as a property wrapper.

mjarvis: Interesting. Some sort of optimization compiler optimization happening. Might be interesting to take a look at some assembled output of a minimal test case to see what its doing.

basememara:
> @basememara There is a very basic dispatch performance test included in project in `ReSwiftTests/PerformanceTests.swift` that may be sufficient for at least the simple time-based comparison.

Awesome thx for pointing that out.

The before and after performance is negligible.. I even watched the CPU and memory and they were identical. Both pegged at 100%; surprisingly the `OSAtomicTestAndClear` didn't go above 100% like its `OS` counterparts go all the way above 1000%. The `GCD` was slightly better at 0.792s and `OSAtomicTestAndClear` was at 0.826, but nothing worth changing for and I'm sure different conditions will be too close to call any a winner.

So the only thing this PR buys us is using GCD over low level `OS` API's.

Before with `OSAtomicTestAndClear`:
Screen Shot 2020-08-18 at 4 02 25 PM

After with `GCD`:
Screen Shot 2020-08-18 at 4 05 41 PM

dani-mp:
If performance is equal or better I'm in favor of changing it for the higher level API. 👏

mjarvis:
I agree as well. Lets get a changelog entry and then once CI is passing we can merge this.

mjarvis:
@basememara I know dealing with all the swift + Xcode variations can be a bit awkward. Let us know if you'd prefer someone else handle clearing that up.

basememara:
Oh phew that would be really amazing pls! I gave a few attempts and I think I'm polluting the commits.

DivineDominion:
Don't worry! We're squashing anyway :) The change to a higher-level API sounds great. Another cool thing to ship with the new release once Xcode 12 is finally out. 🎉

mjarvis (approved):

PR #447, opened 2020-09-19
Resolve iOS deployment target version warning

DominicSchiller-IBM-CIC:
Using the ReSwift library, Xcode permanently shows the following warning at build time:

ios-deployment-target-warning

To resolve this warning, I raised the minimum iOS platform version to `v9` since this is the official least minimum version available to set (according to the error message above).

### DIFF
--- a/Package@swift-5.0.swift
+++ b/Package@swift-5.0.swift
@@ -3,7 +3,7 @@ import PackageDescription
 
 let pkg = Package(name: "ReSwift")
 pkg.platforms = [
-    .macOS(.v10_10), .iOS(.v8), .tvOS(.v9), .watchOS(.v2)
+    .macOS(.v10_10), .iOS(.v9), .tvOS(.v9), .watchOS(.v2)
 ]
 pkg.products = [
     .library(name: "ReSwift", targets: ["ReSwift"])


dani-mp (commented):
It'd be great to change it in the Xcode project as well. Also, do you have info about Apple removing v9 officially?

DominicSchiller-IBM-CIC:
> It'd be great to change it in the Xcode project as well. Also, do you have info about Apple removing v9 officially?

Good point. Because this issue seems to be just related to Apples swift-tools since v5.3 (shipped with Xcode 12), I left the iOS 8 minimum target setting in the Xcode project, so iOS 8 can still be targeted e.g. via CocoaPods or Carthage. 

Unfortunately, I could not find any information yet that Apple is going to drop v9 support officially. On the other hand, the percentage of the [worldwide distribution of iOS versions](https://gs.statcounter.com/os-version-market-share/ios/mobile-tablet/worldwide) < v10.3 is extremely small (~1.66%), so who knows.

gooson (approved):
good to me

gooson (commented):
It's a perfect fix

gooson:
@dani-mp @DominicSchiller-IBM-CIC 

I think it will be solved by modifying the file below.
https://github.com/ReSwift/ReSwift/blob/master/Package.swift

platforms: [**.iOS(.v9)**, .macOS(.v10_11), .tvOS(.v10)],

Other libraries
https://github.com/onevcat/Kingfisher/blob/master/Package.swift
https://github.com/firebase/firebase-ios-sdk/blob/master/Package.swift

dani-mp:
I still think we need to apply this change everywhere: Xcode proj, CocoaPods configuration, SPF configuration. And it'll be a breaking change for projects <9.0, which I understand won't be many, of course, but it will be a breaking change nevertheless. What do you think @mjarvis, @DivineDominion?

mjarvis:
My comment on the associated issue that recently came up:

https://github.com/ReSwift/ReSwift/issues/448#issuecomment-700088255

> This warning is showing up because as of the current release, we still support iOS 8.0 as a valid target, by using an older version of Xcode.
> 
> We are committed to maintaining a sizeable amount of backwards compatibility, but understand this is pushing it a bit far.
> 
> I plan to do another assessment of minimum targets + swift versions before we do a 6.0.0 release (which already drops support for a number of Swift versions)

Basically: I'm okay with us bumping this up as long as we do our due diligence and ensure that its updated *everywhere* in the project. We already have breaking changes in the pipeline for the next release by dropping certain swift versions anyways.

DivineDominion:
I think dropping iOS 8 for the v6.0.0 release is fine. I'm doing the rest in the release branch. Thanks @DominicSchiller-IBM-CIC !

PR #451, opened 2020-10-05
Update docs to denote enum-based actions.

GetToSet:
After adopting ReSwift in some of my projects, I found that combining related actions in enums can result cleaner code which is more readable and maintainable.

This pull request proposes to update related docs to denote this pattern.

### DIFF
--- a/Docs/Getting Started Guide.md
+++ b/Docs/Getting Started Guide.md
@@ -138,6 +138,15 @@ struct SetOAuthURL: Action {
 }
 ```
 
+Or you can combine related actions into an enum type:
+
+```swift
+enum AuthenticationAction: Action {
+    case setOAuthURL(oAuthUrl: URL)
+    case updateLoggedInState(loggedInState: Bool)
+}
+```
+
 ## Reducers
 
 Reducers are the **only place** in which you should **modify** application state. Reducers take the current application state and an action then return the new transformed application state. We recommend to provide many small reducers that each handle a subset of your application state.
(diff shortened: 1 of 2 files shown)


DivineDominion:
These are very valid suggestions! The way you interspersed the info in various places make me wonder: for someone getting started, would it make sense to "refactor" the docs here and make a subsection about the enum-based approach where `Action` declaration and reducer implementation are next to each other?

dani-mp:
I agree with @DivineDominion, but also, although I like the enum-based approach, the Action type is just a protocol, so anyone that knows Swift will also know that you can use other types to conform to that protocol to implement your actions. I'd prefer to keep docs as small and simple as possible.

mjarvis:
I also concur with @dani-mp here. This is a valid approach but in my opinion unnecessary to document.

Another thing to take into account is generally speaking enums should be avoided for "open" (likely to be expanded on) types. Promoting this pattern can result in versioning & breaking changes issues if someone were to publish a public library that exposed actions in this form.
Locally, however, that pattern can be beneficial by forcing implementation of all cases of the switch.

GetToSet:
Thanks for giving suggestions, I'll try to do some refactor, moving related contents to an independent section.

DivineDominion:
@GetToSet Did you have the chance to revise this?

GetToSet:
> @GetToSet Did you have the chance to revise this?

Sorry but currently I have no spare time to further refine these docs. I'll close this pull request.

PR #462, opened 2020-11-10
Deprecate `StateType` protocol requirement

mjarvis:
Improvements in Swift Generics over the years have allowed us to remove this protocol requirement.

As far as I can tell we have no general need for this type, and can usually use `Any` or a generic in place of this when necessary (state types for middleware, etc).

The type has been left in, but deprecated, so this should not be a breaking change.

### DIFF
--- a/ReSwiftTests/StoreMiddlewareTests.swift
+++ b/ReSwiftTests/StoreMiddlewareTests.swift
@@ -9,7 +9,7 @@
 import XCTest
 @testable import ReSwift
 
-let firstMiddleware: Middleware<StateType> = { dispatch, getState in
+let firstMiddleware: Middleware<Any> = { dispatch, getState in
     return { next in
         return { action in
 
@@ -23,7 +23,7 @@ let firstMiddleware: Middleware<StateType> = { dispatch, getState in
     }
 }
 
-let secondMiddleware: Middleware<StateType> = { dispatch, getState in
+let secondMiddleware: Middleware<Any> = { dispatch, getState in
     return { next in
         return { action in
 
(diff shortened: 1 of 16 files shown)


ReSwiftTests/StoreMiddlewareTests.swift
```diff
@@ -9,7 +9,7 @@
 import XCTest
 @testable import ReSwift
 
-let firstMiddleware: Middleware<StateType> = { dispatch, getState in
+let firstMiddleware: Middleware<Any> = { dispatch, getState in
```
dani-mp, line 12: This is the part I don't understand. Would `getState` here know the type of the store's state? If so, does Middleware have to be generic over something at all?

mjarvis: This middleware test doesn't use `getState` -- in middleware that needs to access state, this would be `Middleware<AppState>` or some subset protocol, IE: `struct AppState: HasName`, and then `Middleware<HasName>`.

dani-mp: Alright, makes sense!

DivineDominion (changes requested):
The code changes look good! I wonder why we didn't drop this sooner :) 

We need updates to the docs once this is part of master and people begin to use this. I'd suggest doing both in this PR because of that.

DivineDominion (approved):

DivineDominion:
@mjarvis Sorry, I must've scrolled by the Getting Started.md entry in the file list. I now see that it's there :)

PR #463, opened 2020-11-19
Fix bug where automaticallySkipsRepeats is ignored on generic StoreType

mjarvis:
If a user was to make a subscription to equatable state using a generic `S: StoreType`, swift would use the default implementation in the store rather than the `SubState: Equatable` extension.

By adding this variation to the `StoreType` protocol, we can ensure that the correct function is called.

A unit test has been added that fails before the protocol addition, and passes after.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,10 @@
-<!--
 # Upcoming
 
 **Breaking API Changes:**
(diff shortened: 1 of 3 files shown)


DivineDominion (approved):
Good catch!

PR #467, opened 2020-12-22
Updated Readme

isnine:
Otherwise, it cannot be compiled according to Readme

image


DivineDominion:
Hi @isnine -- thanks for taking the time to take action!  The `StateType` protocol was removed on master after v6.0.0: https://github.com/ReSwift/ReSwift/blob/master/CHANGELOG.md

And I see how that's confusing because the README now doesn't reflect the current state of the library anymore. Hm.

Maybe we should be publishing a new release rather sooner than later to make exploring ReSwift simpler (cc @mjarvis @dani-mp)

PR #473, opened 2021-03-17
When crashing in concurrent acces, log which action caused the error

DivineDominion:
I had one crash report in one of my apps about this, and symbolicating the crash log to figure out what is what in the call stacks kind of sucked :) I think this would've been helpful to more easily trace which action was dispatched that caused the problem.


dani-mp (approved):
💯

mjarvis (approved):

PR #479, opened 2021-06-14
Synchronise access to state for thread-safety

lukeredpath:
Ensures that both read and write access to the store's state is thread-safe by synchronising access over concurrent queue with a barrier as a mutex.

### DIFF
--- a/ReSwift/CoreTypes/Store.swift
+++ b/ReSwift/CoreTypes/Store.swift
@@ -17,17 +17,19 @@ open class Store<State>: StoreType {
 
     typealias SubscriptionType = SubscriptionBox<State>
 
-    private(set) public var state: State! {
+    private(set) var _state: Synchronized<State?> {
         didSet {
             subscriptions.forEach {
                 if $0.subscriber == nil {
                     subscriptions.remove($0)
                 } else {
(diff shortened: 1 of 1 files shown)


mjarvis:
@lukeredpath ReSwift is intended to be single-threaded. What effect does this have on performance? A small change to dispatch/subscription performance makes a big difference when you account for a large application with many actions being dispatched.

lukeredpath:
@mjarvis so I actually meant to open this PR to our own internal fork 🤦🏻 however as it's open I thought I'd leave it open for discussion.

I agree that this is probably best used in a single-threaded way. Unfortunately we have a multi-threaded side-effect system that we're trying to move away from but in the meantime, we have tried to make it thread-safe as we are seeing some strange memory-related crashes in our crash logs.

As implemented this probably isn't the best approach for the core library - perhaps a better alternative is a dedicated Store implementation that can be used concurrently. In terms of performance though, I saw no obvious regression in a performance test that performed 100k actions in a tight loop.

DivineDominion:
I'm closing the PR for now since the overhead of _always_ hopping threads doesn't apply to most apps, and there's no way to opt-in to thread safe access and the performance penalty that's associated with this.

Issue #455, opened 2020-10-19
SwiftUI compatibility plan


mjarvis:
I thought it prudent to start an issue which can describe the necessary steps forward for releasing a SwiftUI compatible version of ReSwift.

At this time, I have a port of ReSwift in [mjarvis/swiftui](https://github.com/ReSwift/ReSwift/tree/mjarvis/swiftui) which conforms `StoreType` to `ObservableObject`. This allows for one to use `@ObservedObject` or `@EnvironmentObject` property wrappers to access a store, allowing direct usage of `store.state.xyz` in view bodies.

There are a number of concerns that need to be taken care of before this branch can be released in a supported manner.

We can split off individual issues and a milestone for completing the work below after #1 is agreed upon.

## 1. Implementation

Is my choice of implementation ideal? Perhaps there are other options than making `StoreType: ObservableObject`.
Maybe something around calling `subscribe` to create a publisher from an injected store?
Do others have suggestions? Or just need to get approvals for existing implementation.

## 2. Backwards-compatibility

We need to determine how we will include this functionality in a general ReSwift release. It is important that we maintain backwards compatibility for `UIKit`, and allow for bug fixes / feature release to continue for both paths.

## 3. Performance

Some general performance testing needs to be done to validate that this is an okay solution for a production environment.
Questions arise such as: How many/often state changes can occur before we overwhelm the SwiftUI diffing? Do we need to do some sort of diffing ourselves before?

## 4. Documentation & Examples

Documentation and examples need to be split and updated to account for UIKit vs SwiftUI paths.

afitterling:
I am quite new to Swift but I am not new to Redux. I used it in other languages. 
Seen from other frameworks and seen from functional reactive programming pattern which combines a iterator and a subscribable it should be fine IMHO. You can always have a look at this: https://redux.js.org/recipes/usage-with-typescript
I have not used the new SwiftUI states... my question is if any other thing like ReactiveSwift RxSwift can be wrapped around the SwiftUI. (https://github.com/ReactiveX/RxSwift) Having it decoupled means creater flexibility and longevity. 
Thanks to you guys I just wrote my first watchKitApp at industry strength.

afitterling:
Rx is a generic abstraction of computation expressed through Observable<Element> interface.
https://github.com/ReactiveX/RxSwift
https://github.com/RxSwiftCommunity/RxState

dani-mp:
[...]

PR #481, opened 2021-07-01
Updates StoreType.swift to have @available annotation

alexwhb:
XCode will not allow my WatchOS project to build without this.

### DIFF
--- a/ReSwift/CoreTypes/Store.swift
+++ b/ReSwift/CoreTypes/Store.swift
@@ -15,6 +15,8 @@ import Combine
  reducers you can combine them by initializng a `MainReducer` with all of your reducers as an
  argument.
  */
+@available(watchOS 6.0, *)
+@available(iOS 13.0, *)
 open class Store<State: StateType>: StoreType, ObservableObject {
 
     @Published private(set) public var state: State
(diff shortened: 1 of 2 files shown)


DivineDominion (changes requested):
Thanks for looking into this! I don't recall this was ever a problem. Which Xcode are you using?

Also, could you pick the lowest possible versions? Should be declared in the project itself, or the .podspec -- e.g. I'm using ReSwift in an app targeting macOS 10.11 no problem, but this change would break that.

alexwhb:
I'm using Xcode 12.5 with swiftUI on watchOS. I'm no longer using pods. I'm using swiftPM which is, I believe, what's causing the discrepancy.

DivineDominion:
@alexwhb Oh sorry, I didn't notice this was merging into @mjarvis's branch! Nevermind then :)

mjarvis:
@alexwhb Hi! In my opinion, this branch is not currently the "best" solution for integrating with SwiftUI.

Please see the discussion in https://github.com/ReSwift/ReSwift/issues/455 -- There are a few lightweight wrappers that require zero changes to the ReSwift source code.

alexwhb:
@mjarvis awesome! I'll check that issue out. Thanks for letting me know.

DivineDominion:
How do you want to proceed with the PR, folks?  Close or change? @alexwhb @mjarvis

DivineDominion:
Am closing due to inactivity.

PR #485, opened 2021-11-27
Fix test for intial state dispatch upon subscription

DivineDominion:
I noticed that the test for initial state dispatching to subscribers contained a state change via `SetValueAction`.

So the test didn't actually assert what it said.

I split this into 3 tests: one for simple state change, two for initial state dispatching.

This is such a weird oversight that I wonder if I interpreted the tests wrong, though.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 **API Changes:**
 
 **Other:**
+- Add tests to clarify initial state dispatch (#485) - @DivineDominion
 
 # 6.1.0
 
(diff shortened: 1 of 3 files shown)


dani-mp (approved):

DivineDominion:
Oh no, forgot Travis CI is broken here as well :|

mjarvis (approved):
You interpreted this correctly. Nice catch!

mjarvis:
I have disabled the branch protection requirement for Travis for the time being. We will need to be extremely diligent that everything compiles + tests pass before merging.

This PR is 👍  because it is only in test code. No actual framework code affected.

DivineDominion:
Was hesitant to disable protection, but it's probably better than not making any progress until we add GitHub Actions to this repo. -- If any one of you has an idea how we can make the multi-platform and version testing better than the abomination I came up with in ReSwift-Thunk, please do share :) https://github.com/ReSwift/ReSwift-Thunk/blob/master/.github/workflows/cross-platform-compile.yml

PR #486, opened 2022-01-07
Migrate to Github Actions

### LINKED ISSUE #484: Remove Travis from all ReSwift repos? / How to set up GH Actions
Broken Travis CI is getting in the way of reviews and doesn't add value anymore. I'd say we remove it ASAP.

But what do we use instead?

I had a hard time figuring out how to get GH Actions into some kind of semi-working condition. I really had no clue what I was doing, and feedback would be much appreciated. Maybe we can then more or less  copy the setup instructions over.

https://github.com/ReSwift/ReSwift-Thunk/pull/53

mjarvis:
Resolves #484, #486 

Migrates CI to GitHub actions.

TODO:
- [x] Update `.scripts/update-gh-pages` to work with github actions


DivineDominion (approved):
This looks pretty good and the actions do run, so nice work! 

Just looking at the result makes it seem so easy. But the commit history reveals you had to battle with the system, too. I feel you, thinking back to the ReSwift-Thunk GH Action implementation I somehow forced to do something useful :)

(I guess caching would be nice to both speed up the process and of course preserve energy to help fight climate change etc. ;) but we don't run CI that much, so I think this is pretty good.)

Your initial PR comment has an unticked box w.r.t. documentation generation. Doesn't the "Update Docs" workflow now already do this? Or does it just call the script but the script isn't doing it properly?

mjarvis:
@DivineDominion The update docs script appears to be heavily dependent on travis-ci environment variables. Haven't looked closely but I'm assuming it will need modifications. Will take a look today.

.scripts/update-gh-pages
```diff
@@ -71,12 +56,12 @@ pushd "$CHECKOUT_PATH"
...
     git add -A "$BRANCH"
-
+    
     echo " ==> Commit changes"
     git commit -m "[$BRANCH $SHA] Regenerate documentation"
-
+    
     echo " ==> Push changes -> '$REPO'"
-    git push -q "https://$GITHUB_PAGES_TOKEN@github.com/$ORGANISATION/$NAME.git" gh-pages
+    git push -q "https://github-actions:$GITHUB_TOKEN@github.com/$ORGANISATION/$NAME.git" gh-pages
```
DivineDominion, line 64: That's a good find! 👍

.scripts/update-gh-pages
```diff
@@ -36,24 +27,18 @@ else
     git clone --branch gh-pages "$REPO" "$CHECKOUT_PATH"
 fi
 
-if [ "$TRAVIS_BRANCH" != "" ]; then
-    echo " ==> Install: jazzy"
-    gem install --no-rdoc --no-ri jazzy
```
DivineDominion, line 41: Sorry if that's a dumb question: How's the binary dependency resolved now if we skip the installation step?

DivineDominion (approved):
Looks good in principle, but I'd still be curious to see if this works :)

.scripts/update-gh-pages
mjarvis: Not dumb. Jazzy is pre-installed as per https://github.com/actions/virtual-environments/blob/main/images/macos/macos-11-Readme.md

PR #491, opened 2022-09-14
Replaced open func subscribe<SelectedState: Equatable, S: StoreSubscr…

maksimgromov:
…iber> with public func subscribe<SelectedState: Equatable, S: StoreSubscriber> and open func subscribe<S: StoreSubscriber>(_ subscriber: S) with public func subscribe<S: StoreSubscriber>(_ subscriber: S) as required.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 **API Changes:**
 
 **Other:**
+- Replaced open func subscribe<SelectedState: Equatable, S: StoreSubscriber> with public func subscribe<SelectedState: Equatable, S: StoreSubscriber> and open func subscribe<S: StoreSubscriber>(_ subscriber: S) with public func subscribe<S: StoreSubscriber>(_ subscriber: S) since non-'@objc' instance method in extensions cannot be overridden, should use 'public' instead (#491) - @maksimgromov
 - Add tests to clarify initial state dispatch (#485) - @DivineDominion
 
 # 6.1.0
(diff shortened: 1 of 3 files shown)


DivineDominion (changes requested):
Thanks!

Could you add what your motivation for this change was? (Hmm, can't override functions from extensions anyway, so the `open` modifier wasn't helping much, I guess.)

Please add an entry to the CHANGELOG file with a short summary 🙏 

I went ahead and removed the hidden Xcode workspace file from the PR.

maksimgromov:
Hello. I made this changes with respect to compiling advice during fastlane gym procedure.
Screenshot 2022-09-16 at 11 57 34

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 **API Changes:**
 
 **Other:**
+- Replaced open func subscribe<SelectedState: Equatable, S: StoreSubscriber> with public func subscribe<SelectedState: Equatable, S: StoreSubscriber> and open func subscribe<S: StoreSubscriber>(_ subscriber: S) with public func subscribe<S: StoreSubscriber>(_ subscriber: S) since non-'@objc' instance method in extensions cannot be overridden, should use 'public' instead (#491) - @maksimgromov
```
DivineDominion, line 8: What do you think about the following to "hide" the details and focus on the intention?

```suggestion
- Replaced `open func` with `public func` in extensions to `Store` because they cannot be overridden anyway (#491) - @maksimgromov
```

DivineDominion (changes requested):
👍 Makes sense, thanks!

DivineDominion (approved):
Looks good, thanks!

PR #502, opened 2023-01-06
Add Carthage instructions to build a binary release

DivineDominion:
With `.xcframeworks`, it's not as easy as it used to be.


mjarvis (approved):

DivineDominion:
Thanks for taking a look so quickly, @mjarvis!

PR #503, opened 2023-04-18
New Subscribe Methods Using AnySubscriber as an Arg

jeffhandy:
I've never liked all the boilerplate code for multi-Subscriptions--ie. subscriptions instantiated in a viewController that implement StoreSubscriber. I've managed to code a solution that would allow devs to use a shortcut notation using a keyPath like so:

```
handleSubscription( \.deviceState.applicationState ) { [weak self] state in
    if let error = state as? ApplicationState {
          // do work here
     }
   }
```

versus the standard way:

```
private lazy var applicationStateSubscriber
     = AnonymousStoreSubscriber<DeviceState> { [weak self] state in
       self?.newApplicationState(in: state)
     }
   
   store.subscribe(applicationStateSubscriber) { subscription in
     subscription
       .select { $0.deviceState }
       .skip { $0.applicationState == $1.applicationState }
   }
```
   
By implementing an `AnySubscriber` like this:
   
```
public class ManySubscriber: AnyStoreSubscriber {

    public init(onNewState: @escaping (Any) -> Void) {
        self.onNewState = onNewState
    }

    public func _newState(state: Any) {
        self.onNewState(state)
    }

    public var onNewState: (Any) -> Void
}
```

We can add this `ManySubscriber` to a base viewController for managing subscriptions :

  ```
   var subscribers = [ManySubscriber]()

    func handleSubscription<SelectedState: Equatable>(_ keyPath: WritableKeyPath<AppState, SelectedState>,
        onNewState: @escaping (Any) -> Void ) {

        let subscriber = ManySubscriber { _ in }
        subscribers.append(subscriber)
        subscriber.onNewState = onNewState
        store.subscribeAny(subscriber, keyPath: keyPath)
    }

    deinit {
        subscribers.forEach { store.unsubscribe($0) }
    }
```

I've intentionally called the new method `subscribeAny()` to avoid confusion with the original `subscribe()` method. 

I am aware that this approach means the Value passed in the KeyPath will not be **_**typed**_**.
   

**UPDATE:**  I've added a second overloaded `subscribeAny()` call which takes 2 closures as args--a subscription and newState handler. It is invoked liked so:

```
handleSubscription( { subscription in
          subscription
              .select { $0.state1 }
              .skip { $0.counter1 == $1.counter1 }
      }) { [weak self] state in
          if let state = state as? State1 {
              self?.counterLabel.text = "\(state.counter1)"
          }
      }
```
      
In this call we conveniently keep together in one place WHAT WE ARE SUBSCRIBING TO and HOW WE HANDLE IT.

### DIFF
--- a/ReSwift/CoreTypes/Store.swift
+++ b/ReSwift/CoreTypes/Store.swift
@@ -246,27 +246,49 @@ extension Store where State: Equatable {
 // MARK: Allow subscriptions using Any, so State not strongly typed
 
 extension Store {
-    
    public func subscribeAny<S: AnyStoreSubscriber, SelectedState: Equatable>(
-    _ subscriber: S, keyPath: WritableKeyPath<State, SelectedState>, skipFirstState: Bool = false)
-    {
+    _ subscriber: S, keyPath: WritableKeyPath<State, SelectedState>, skipFirstState: Bool = false) {
+        
         // Create a subscription for the new subscriber.
         let originalSubscription = Subscription<State>()
-        
-        // Transform by selecting a subState with keyPath
+
+        // Transform by selecting a subState using the keyPath arg
         var transformedSubscription = originalSubscription.select(keyPath)
+
+        if subscriptionsAutomaticallySkipRepeats {
+            transformedSubscription = transformedSubscription.skipRepeats()
+        }
+
         if skipFirstState {
             transformedSubscription = transformedSubscription.skipFirstState()
         }
-        transformedSubscription = transformedSubscription.skipRepeats()
+
         _subscribeAny(subscriber, originalSubscription: originalSubscription,
-                   transformedSubscription: transformedSubscription)
+                      transformedSubscription: transformedSubscription)
     }
+    
+    public func subscribeAny<S: AnyStoreSubscriber, SelectedState: Equatable>(
+        _ subscriber: S, transform: ((Subscription<State>) -> Subscription<SelectedState>)?) {
+            
+         // Create a subscription for the new subscriber.
+         let originalSubscription = Subscription<State>()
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by mjarvis** [comment]
Hi Jeff, this is a common extension that is used for multiple subscribers, and can be achieved in a type safe manner using a type such as `BlockSubscriber`. Here is a simple set of extensions that you can use in your projects:

We've considered adding this sort of thing to the project but at the moment we're holding off. Its simple enough to include yourself when needed.

```swift
class BlockSubscriber<S>: StoreSubscriber {
    private let block: (S) -> Void

    init(block: @escaping (S) -> Void) {
        self.block = block
    }

    func newState(state: S) {
       block(state)
    }
}

extension StoreType {
    func subscribe<SubState>(
        select: @escaping (State) -> SubState,
        newState: @escaping (SubState) -> Void
    ) -> BlockSubscriber<SubState> {
        let subscriber = BlockSubscriber(block: newState)
        subscribe(subscriber) { $0.select(select) }
        return subscriber
    }

    func subscribe<SubState: Equatable>(
        select: @escaping (State) -> SubState,
        newState: @escaping (SubState) -> Void
    ) -> BlockSubscriber<SubState> {
        let subscriber = BlockSubscriber(block: newState)
        subscribe(subscriber) { $0.select(select) }
        return subscriber
    }
}

class Foo {
    private var subscribers: [AnyStoreSubscriber] = []

    init(store: Store<Bar>) {
        subscribers = [
            store.subscribe(select: \.someState) { state in
                print(state)
            },
            store.subscribe(select: \.otherState) { state in
                print(state)
            }
        ]
    }
}
```

PR #504, opened 2023-05-01
Update GitHub actions with new Xcode select action

DivineDominion:
`devbotsxyz/xcode-select` is archived, so I switched to an action that's still supported: https://github.com/marketplace/actions/xcode-select-version


DivineDominion:
We will likely need to adjust the Test Matrix for Xcode 14. 

I'm working on that in a parallel branch.

Problem with that is: the macOS version bump from `macos-11` to `macos-13` makes the Swift 5.2 backwards compatibility build obnoxious :) Need to put the macOS version into the test matrix, too, maybe. -- Or we remove that one 👀

DivineDominion:
@mjarvis Do you have any idea why the iOS test might fail? Asking first to avoid one or more hours of research -- but if you also don't know, no problem :)

mjarvis:
> @mjarvis Do you have any idea why the iOS test might fail? Asking first to avoid one or more hours of research -- but if you also don't know, no problem :)

Not sure -- I've re-run the job to see if it was just a temporary issue. The `xcodebuild` command works for me locally.

DivineDominion:
That's weird.

I've triggered another build with debug logging to see if there's a difference.

This is terrible, because then the CI step is useless.

DivineDominion:
GH Actions broke and merging was impossible

PR #510, opened 2023-10-09
Bump deployment targets, remove old versions from CI

DivineDominion:
Another try. Couldn't get the old stuff to work. I figured we might as well require people to pin 6.1.1 if they reaaaaally need Swift 4.x compilation support.


DivineDominion:
@mjarvis  Sorry for the squashed commit. I didn't understand why merging was disabled and tried to change the PR history from others to maybe invalidate a cache -- well turns out I needed to update the master branch protection rules to use the new test names!

TLDR:  dropped Swift 4.x support, bumped all deployment targets to compile with Xcode 14.2

mjarvis (approved):
Sorry for the huge delays in response, @DivineDominion, thanks for updating this!

